PR #62, opened 2015-02-04
Add 'org.jcp.xml.dsig.internal' as system package.

xiaodong-xie:
1. This package is included in both jdk7 and jdk8.
2. Use package will be used as a typical jsr105 usage, such as 
```
        String providerName = System.getProperty("jsr105Provider", "org.jcp.xml.dsig.internal.dom.XMLDSigRI");
        XMLSignatureFactory fac =
            XMLSignatureFactory.getInstance("DOM", (Provider) Class.forName(providerName).newInstance());
```

### DIFF
--- a/src/main/java/org/jboss/modules/Module.java
+++ b/src/main/java/org/jboss/modules/Module.java
@@ -88,6 +88,7 @@ public final class Module {
         final List<String> list = new ArrayList<String>();
         list.add("java.");
         list.add("sun.reflect.");
+        list.add("org.jcp.xml.dsig.internal");
         list.add("__redirected.");
         if (pkgsString != null) {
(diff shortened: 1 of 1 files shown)


dmlloyd:
This is not the correct fix.  None of the other JAXP implementation classes are listed as system packages; this should be fixed in the JAXP redirection code (though I do not know why it has failed in the first place).

xiaodong-xie:
Thanks, @dmlloyd. :+1: Problem solved.

xiaodong-xie:
Just use something like 
```
XMLSignatureFactory fac = XMLSignatureFactory.getInstance("DOM", "XMLDSig");
```
is enough.

PR #67, opened 2015-06-18
Expose mxBean via public api

ctomc:


ctomc:
not really that big of a deal

PR #68, opened 2015-06-23
Remove JAXP redirect hacks, as JAXP 1.6 that is part of JDK8 works be…

ctomc:
…tter in modular environments


src/main/java/org/jboss/modules/Main.java
```diff
@@ -373,11 +368,11 @@ public static void main(String[] args) throws Throwable {
             moduleIdentifier = ModuleIdentifier.fromString(nameArgument);
         }
         Module.initBootModuleLoader(loader);
-        if (jaxpModuleIdentifier != null) {
```
dmlloyd: Please remove this instead of commenting it out.

src/test/java/org/jboss/modules/JAXPModuleTest.java
```diff
@@ -210,7 +193,7 @@ public void setupModuleLoader() throws Exception {
             throw new IllegalStateException(e);
         }
     }
-
+  /*
```
dmlloyd: Again please remove.

src/main/java/org/jboss/modules/Main.java
```diff
@@ -35,18 +33,17 @@
...
-import java.util.logging.LogManager;
-
 import java.util.jar.Manifest;
+import java.util.logging.LogManager;
 import java.util.prefs.Preferences;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import org.jboss.modules.log.JDKModuleLogger;
-
```
dmlloyd: Reordered imports.

PR #72, opened 2015-08-04
Close InputStream after reading it

gastaldi:
@dmlloyd Can you review it?

### DIFF
--- a/src/main/java/org/jboss/modules/FileResourceLoader.java
+++ b/src/main/java/org/jboss/modules/FileResourceLoader.java
@@ -94,12 +94,16 @@ final class FileResourceLoader extends NativeLibraryResourceLoader implements It
     }
 
     private static Manifest readManifestFile(final File manifestFile) {
-        try {
-            return manifestFile.exists() ? new Manifest(new FileInputStream(manifestFile)) : null;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/modules/FileResourceLoader.java
```diff
@@ -94,11 +94,18 @@ public URL run() {
     }
 
     private static Manifest readManifestFile(final File manifestFile) {
+        Manifest manifest = null;
+        FileInputStream fis = null;
         try {
-            return manifestFile.exists() ? new Manifest(new FileInputStream(manifestFile)) : null;
```
dmlloyd: Instead maybe just use try-with-resources?

gastaldi: Is it being compiled with JDK7 ? I missed that.

gastaldi: Try-with-resources still asks to catch IOException. I think this is fine

gastaldi:
@dmlloyd PR updated, please review. I wasn't sure about the code formatter used, so I used the Java Conventions (Built-in) in Eclipse

PR #75, opened 2015-09-23
[MODULES-217] more robust PathUtils.is*Child() implementations

ropalka:


dmlloyd:
Can you do one for master as well please?

PR #79, opened 2015-11-12
put a null guard around PropertyWriteAction

djeremiah:
classPath may be null when -class option is specified in Main and CLASSPATH environment variable is undefined.

This is the most expedient way to deal with it, but I'm not sure it's the best. 
Other options:
1) remove PropertyWriteAction entirely
java.class.path is already set by Main if -classpath option is specified but we should probably honor 
CLASSPATH for -class option.
2) provide a sensible default
replace line 44 with System.getenv().getOrDefault(). Though I'm not sure what a sensible, non-null default is for java.class.path. Is "" a valid value?

### DIFF
--- a/src/main/java/org/jboss/modules/ClassPathModuleLoader.java
+++ b/src/main/java/org/jboss/modules/ClassPathModuleLoader.java
@@ -41,12 +41,10 @@ final class ClassPathModuleLoader extends ModuleLoader {
         this.delegateLoader = delegateLoader;
         this.dependencies = dependencies;
         if (isEmpty(classPath)) {
-            classPath = System.getenv().get("CLASSPATH");
+            classPath = System.getenv().getOrDefault("CLASSPATH","");
         }
         this.classPath = classPath;
-        if(classPath != null){
-        	AccessController.doPrivileged(new PropertyWriteAction("java.class.path", classPath));
-        }
+       	AccessController.doPrivileged(new PropertyWriteAction("java.class.path", classPath));
         this.mainClass = mainClass;
     }
 
--- a/src/main/java/org/jboss/modules/ClassPathModuleLoader.java
+++ b/src/main/java/org/jboss/modules/ClassPathModuleLoader.java
@@ -44,7 +44,9 @@ final class ClassPathModuleLoader extends ModuleLoader {
             classPath = System.getenv().get("CLASSPATH");
         }
         this.classPath = classPath;
-        AccessController.doPrivileged(new PropertyWriteAction("java.class.path", classPath));
+        if(classPath != null){
+        	AccessController.doPrivileged(new PropertyWriteAction("java.class.path", classPath));
+        }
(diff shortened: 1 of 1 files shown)


dmlloyd:
I think "" is a valid value for java.class.path.  One way to test that would be to create a test JAR which prints out that property, and then run it as an agent with nothing on the classpath.

djeremiah:
It looks like, at the very least, the Agent.jar is put on the classpath, as well as the jar specified by -jar.

I cooked up an Agent that just prints the property:
```
public static void premain(String agentargs, Instrumentation inst){
		String path = System.getProperty("java.class.path");
		System.out.println("Path: " + path);
}
```
Re running the use original use case from the JIRA:

<pre>
[jboss-eap-6.3]$  java -javaagent:/home/dmurphy/workspaces/jboss-modules/AgentTest/Agent.jar -jar jboss-modules.jar -mp modules/ -dep org.picketbox -class org.picketbox.datasource.security.SecureIdentityLoginModule testString

<strong>Path: jboss-modules.jar:/home/dmurphy/workspaces/jboss-modules/AgentTest/Agent.jar</strong>
Exception in thread "main" java.lang.NullPointerException
    ...etc...
</pre>

So I'm not sure about using "" for the default.

dmlloyd:
Ah, I figured that using an agent would put it on the bootstrap class path instead of the application class path.  Also what I meant was that the test program could just run on a raw JVM with no other JARs on the class path.

But perhaps a better test is to run the jar by appending it directly to the boot class path and then running the class by name, again with an empty class path.  I may run this test yet tonight if I get a minute.

dmlloyd:
Figured out a couple interesting facts:

* The default class path is . - that's one of those facts you learn in your first month of Java and then forget immediately thereafter :-)
* If you use the bootclasspath hack, you can manually set the class path to empty string

So I think that empty string is the best default.

dmlloyd:
At least it's better than leaving it unset to default to ????? whatever it might be defaulting to.

djeremiah:
Makes sense to me. PR updated.

src/main/java/org/jboss/modules/ClassPathModuleLoader.java
```diff
@@ -41,10 +41,10 @@
         this.delegateLoader = delegateLoader;
         this.dependencies = dependencies;
         if (isEmpty(classPath)) {
-            classPath = System.getenv().get("CLASSPATH");
+            classPath = System.getenv().getOrDefault("CLASSPATH","");
         }
         this.classPath = classPath;
-        AccessController.doPrivileged(new PropertyWriteAction("java.class.path", classPath));
+       	AccessController.doPrivileged(new PropertyWriteAction("java.class.path", classPath));
```
dmlloyd: Looks like a tab snuck in here?

dmlloyd:
I squashed it all down to the single one-line diff commit and merged it.  Thanks.

PR #80, opened 2015-12-11
JDK9 changes

ctomc:


dmlloyd:
I'm going to close this as 1.5 has better forms of these changes; we should just move to that branch at this point.

PR #83, opened 2015-12-15
MODULES-231 Add checkstyle to build

ctomc:


dmlloyd:
I don't want to include src/main/java/org/jboss/modules/xml/MXParser.java in checkstyle.  Also I'd like the removal of the unused class to be a separate commit to help clarify the impact of this.

ctomc:
updated, I will send separate pull for removal of the class

PR #86, opened 2016-01-08
added support for property expansion in permissions/grant/name attribute

jasinner:
https://issues.jboss.org/browse/MODULES-236

### DIFF
--- a/src/main/java/org/jboss/modules/PropertyExpander.java
+++ b/src/main/java/org/jboss/modules/PropertyExpander.java
@@ -1,0 +1,196 @@
+package org.jboss.modules;
+
+import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.security.GeneralSecurityException;
+import java.util.BitSet;
+
+//Copied from sun.net.www.ParseUtil, and sun.security.util.PropertyExpander
+
+public class PropertyExpander {
+    static BitSet encodedInPath;
+
+    static {
+        encodedInPath = new BitSet(256);
+
+        // Set the bits corresponding to characters that are encoded in the
+        // path component of a URI.
+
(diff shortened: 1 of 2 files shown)


dmlloyd:
This code was copied without copyright or license headers, and it's not clear what the license or copyright status of this code is.  Furthermore, it's not clear that we are legally authorized to relicense the code as ASL.  If this code came from OpenJDK, we definitely are not, since the license of that code (modified GPL + CDDL) does not allow us to relicense the code as ASL as far as I'm aware.

jasinner:
Can we use org.jboss.dmr.ValueExpressionResolver instead of the JDK code?
If so, can we add a dependency from jboss-modules to org.jboss.dmr, or do you want to keep it pure, and just copy in the ValueExpressionResolver minus the code we don't need?

dmlloyd:
It is impossible to add dependencies from jboss-modules to other projects.  So the code would have to be copied, but then stripped down and probably modified to match all the rules defined in the [Oracle specification] (https://docs.oracle.com/javase/8/docs/technotes/guides/security/PolicyFiles.html#PropertyExp).

dmlloyd:
Also the code should be non-public, and I think it should be a part of the XML parsing, not the permission factories themselves.

dmlloyd:
This is ok to test

dmlloyd:
(that's just to satisfy the CI bot)

jasinner:
@dmlloyd Hi David,
I think you are requesting some changes, and I am happy to make changes. However I don't understand what I need to change. Is it possible for you to make the changes yourself?

dmlloyd:
Not at the moment, though I can possibly look into it if/when things calm down.

dmlloyd:
This was replaced by #94.

PR #97, opened 2016-05-17
Implement NIO2 Path-based resource loader

bsideup:
Hi.

Currently, JBoss modules are using old, `java.io.File`-based FileResourceLoader and JarResourceLoader.

Since NIO2 was released, there is really nice FileSystem abstraction. This PR makes it possible to load resources from `Path`, which might be represented by different types of FileSystems, like:

LocalFileSystem (test included)
JarFileSystem (see the test as well)
Arquillian's ShrinkWrap since it implements FileSystem SPI
InMemoryFileSystems like http://github.com/google/jimfs
Some distributed file systems

Generally, I'm going to use PathResourceLoader in Forge's Furnace because I want to distribute Furnace.

### DIFF
--- a/src/main/java/org/jboss/modules/PathResource.java
+++ b/src/main/java/org/jboss/modules/PathResource.java
@@ -1,0 +1,67 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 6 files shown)


dmlloyd:
The change looks good, thanks!  I need to ask this before merging it though:

We require that all contributions be made under the terms of the MIT License, http://www.opensource.org/licenses/mit-license.php. Do you agree to the these terms?

bsideup:
@dmlloyd absolutely agree 👍

dmlloyd:
Great! Thanks.

bsideup:
@dmlloyd right now it looks like it's possible to rewrite FileResourceLoader and JarFileResourceLoader and just use `File#toPath()`, but there are some security checks. Question - can you help me a bit with some context about them? Is it still relevant? If so, I think I can proxy calls to keep these checks.

dmlloyd:
You would definitely need to do the same security procedure that FileEntryResource does, where the AccessControlContext is captured and used in the ``getSize()`` and ``openStream()`` methods with a ``doPrivileged()`` call.

bsideup:
@dmlloyd may I ask you to help me with a test in AbstractResourceLoaderTestCase? Because currently I can't find any tests for the it, and security sounds like a big deal, so it would be nice to have such test :)

dmlloyd:
It's tricky because you need to have a security manager installed to test the security stuff.  I don't have a framework together to do that yet, so I rely on correctness reasoning for now.

bsideup:
Got it. Will try to keep all security checks as much as possible then.

PR #98, opened 2016-05-17
Make FileResourceLoader extends PathResourceLoader and implement security checks

bsideup:
Since #97 was merged, `FileResourceLoader` is no more and could be replaced with `PathResourceLoader` + `File#toPath()`. Only one thing left in FileResourceLoader - index caching, which IMO shouldn't be a part of PathResourceLoader because NIO2 FS might be read-only.

I agree to the terms of the MIT license.

### DIFF
--- a/src/main/java/org/jboss/modules/FileResourceLoader.java
+++ b/src/main/java/org/jboss/modules/FileResourceLoader.java
@@ -18,357 +18,54 @@
 
 package org.jboss.modules;
 
-import static java.security.AccessController.doPrivileged;
-
-import java.io.BufferedReader;
-import java.io.BufferedWriter;
 import java.io.File;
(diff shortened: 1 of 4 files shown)


bsideup:
@dmlloyd I hope it's fine that security checks were implemented in the same PR where I replaced FileResourceLoader with PathResourceLoader, one was required for another :)

bsideup:
@dmlloyd thanks for merging!

dmlloyd:
No problem.  I added a cleanup commit afterwards that you might want to check out.

https://github.com/jboss-modules/jboss-modules/commit/e0c97173f38237ad02265146802155acd9a7e996

bsideup:
@dmlloyd ok, cool, now it's cleaner :)

PR #101, opened 2016-07-05
Code update and version class support

dmlloyd:


src/main/java/org/jboss/modules/ConcurrentClassLoader.java
```diff
@@ -363,7 +364,7 @@ public final InputStream getResourceAsStream(final String name) {
         return findClass(className, exportsOnly, resolve);
     }
 
-    private final UnlockedReadHashMap<String, Package> packages = new UnlockedReadHashMap<String, Package>(64);
+    private final ConcurrentHashMap<String, Package> packages = new ConcurrentHashMap<>();
```
ropalka: I like this clean up.

PR #109, opened 2016-09-14
JDK9 related fixes

ctomc:


src/main/java/org/jboss/modules/JDKPaths.java
```diff
@@ -55,6 +79,51 @@
     private JDKPaths() {
     }
 
+    private static void processRuntimeImages(final Set<String> jarSet) {
+        try {
+            for (final Path root : FileSystems.getFileSystem(new URI("jrt:/")).getRootDirectories()) {
+                Files.walkFileTree(root, new JrtFileVisitor(jarSet));
```
dmlloyd: Note that this will pull in all modules, not just the ones that are actually active/visible at run time.  Might or might not be a problem, but we can revisit later.

PR #111, opened 2016-10-07
[MODULES-270] fixing problem where ClassLoader.definePackage() can th…

ropalka:
…row IllegalArgumentException on JDK9

### DIFF
--- a/src/main/java/org/jboss/modules/ConcurrentClassLoader.java
+++ b/src/main/java/org/jboss/modules/ConcurrentClassLoader.java
@@ -460,6 +460,7 @@ public abstract class ConcurrentClassLoader extends ClassLoader {
                 pkg = super.definePackage(name, specTitle, specVersion, specVendor, implTitle, implVersion, implVendor, sealBase);
             } catch (final IllegalArgumentException iae) {
                 pkg = super.getPackage(name);
+                if (pkg == null) throw iae;
             }
             existing = packages.putIfAbsent(name, pkg);
             return existing != null ? existing : pkg;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/modules/ConcurrentClassLoader.java
```diff
@@ -455,7 +455,12 @@ protected Package definePackage(final String name, final String specTitle, final
             if (existing != null) {
                 return existing;
             }
-            Package pkg = super.definePackage(name, specTitle, specVersion, specVendor, implTitle, implVersion, implVendor, sealBase);
+            Package pkg;
+            try {
+                pkg = super.definePackage(name, specTitle, specVersion, specVendor, implTitle, implVersion, implVendor, sealBase);
+            } catch (final IllegalArgumentException iae) {
+                pkg = super.getPackage(name);
+            }
```
dmlloyd: When does this occur?  Is this something new in Java 9?

ropalka: I noticed that on JDK8 whole method ClassLoader.definePackage() was synchronized.
In JDK9 they're using ConcurrentHashMap.putIfAbsent() instead so it opens a window
for potential concurrent update of the underlying package map. I think this design change is the reason why we can see IllegalArgumentException in some tests.

dmlloyd: Is there any chance an IllegalArgumentException can be thrown without the package being defined in the superclass?

dmlloyd: I think we might want to add an ``if (pkg == null) throw iae;" after that getPackage call.

ropalka: Sure, reworking ...

ropalka: Done.

PR #113, opened 2016-10-26
Commits for JDK 9 support

dmlloyd:


pom.xml
```diff
@@ -43,6 +43,8 @@
     <properties>
         <skip.compile>false</skip.compile>
         <version.org.wildfly.checkstyle-config>1.0.5.Final</version.org.wildfly.checkstyle-config>
+        <maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>1.8</maven.compiler.target>
```
ctomc: redundant as jboss-parent already defaults to 1.8

dmlloyd: True, but in this case I want to be explicit due to the mixed support for 8 and 9 that the main code base is Java 8 only, even though it includes 9.  Also this will help if/when we bump the default to 9 at some future date, I suppose.

src/main/java/org/jboss/modules/ConcurrentClassLoader.java
```diff
@@ -66,17 +66,37 @@
...
+    /**
+     * Construct a new instance with the given parent class loader, which must be a concurrent class loader, or {@code null}
+     * to create a root concurrent class loader.
+     *
+     * @param parent the parent class loader
+     * @param name the name of this class loader, or {@code null} if it is unnamed
+     */
+    protected ConcurrentClassLoader(final ConcurrentClassLoader parent, final String name) {
+        super(parent == null ? ConcurrentClassLoader.class.getClassLoader() : parent, name);
         if (getClassLoadingLock("$TEST$") == this) {
```
ropalka: This check should be:
! JDKSpecific.isParallelCapable(this)

src/main/java/org/jboss/modules/ModuleLoader.java
```diff
@@ -239,6 +239,19 @@ public String toString() {
...
+     * Get a string representation of the given module, used in debug output and stack traces.  This method may be
+     * overridden to provide a more detailed description.  By default it returns a string consisting of the module's
+     * name and version (if any).
+     *
+     * @param module the module to describe
+     * @return the description string
+     */
+    public String getModuleDescription(final Module module) {
+        final Version version = module.getVersion();
+        return version == null ? module.getName() : module.getName() + "@" + version;
```
ropalka: I like this change. This new feature is very helpful to have module names in stack traces.
I already had an opportunity to see that feature in practice while debugging your ModuleIdentifier deprecation changes ;)

pom.xml
```diff
@@ -155,6 +183,18 @@
     </build>
     <dependencies>
         <dependency>
+            <groupId>org.jboss.modules</groupId>
+            <artifactId>jboss-modules-jdk9-supplement</artifactId>
+            <version>1.0.Final</version>
```
ropalka: Ah, you already released it :( I have a couple of fixes for it locally.

src/main/java/org/jboss/modules/JDKPaths.java
```diff
@@ -57,22 +57,23 @@
             moduleClass = JDKPaths.class.getClassLoader().loadClass(JDK9_CLASS);
         } catch (final Throwable ignored) {}
         isJDK9orAbove = moduleClass != null;
-        final Set<String> pathSet = new FastCopyHashSet<>(1024);
+        final Set<String> pathSet;
         final Set<String> jarSet = new FastCopyHashSet<>(1024);
         if (isJDK9orAbove) {
+            // TODO: Remove this to the JDK9 supplement
```
ropalka: This whole 'isJDK9orAbove' branch can be eliminated.
I see it is already present in jboss-modules-jdk9-supplement.

pom.xml
dmlloyd: That's OK, tags are easy to do.  I expected that we may have lots of releases at first.

src/main/java/org/jboss/modules/ConcurrentClassLoader.java
dmlloyd: You're right!  Somehow I missed that.

src/main/java/org/jboss/modules/JDKPaths.java
dmlloyd: Will remove it.

ropalka:
+1 for merge

PR #115, opened 2016-11-04
[MODULES-256] Support slotless module finder path scheme

dmlloyd:


src/main/java/org/jboss/modules/JarModuleFinder.java
```diff
@@ -171,4 +173,13 @@ public ModuleSpec findModule(final String name, final ModuleLoader delegateLoade
             return moduleSpec;
         }
     }
+
+    private static String toPathString(String moduleName) {
+        return moduleName.replace('.', '/') + '/' + moduleName + '/';
```
ropalka: Method body should be: 

return moduleName.replace('.', '/');

without ending slash, because later this path is used in combination with "/module.xml" string.

src/main/java/org/jboss/modules/JarModuleFinder.java
```diff
@@ -171,4 +173,13 @@ public ModuleSpec findModule(final String name, final ModuleLoader delegateLoade
...
+    }
+
+    private static String toLegacyPathString(String moduleName) {
+        final ModuleIdentifier moduleIdentifier = ModuleIdentifier.fromString(moduleName);
+        return moduleIdentifier.getName().replace('.', '/') + '/' + moduleIdentifier.getSlot() + '/';
```
ropalka: Same problem with ending slash here.

src/main/java/org/jboss/modules/LocalModuleFinder.java
```diff
@@ -129,41 +129,37 @@ public LocalModuleFinder(boolean supportLayersAndAddOns) {
         return files;
     }
 
-    private static String toPathString(ModuleIdentifier moduleIdentifier) {
-        final StringBuilder builder = new StringBuilder(40);
-        builder.append(moduleIdentifier.getName().replace('.', File.separatorChar));
-        builder.append(File.separatorChar).append(moduleIdentifier.getSlot());
+    private static String toPathString(String moduleName) {
+        return moduleName.replace('.', File.separatorChar) + File.separatorChar + moduleName + File.separatorChar;
```
ropalka: Similar to above i.e. method body should be:

return moduleName.replace('.', File.separatorChar);

src/main/java/org/jboss/modules/LocalModuleFinder.java
```diff
@@ -129,41 +129,37 @@ public LocalModuleFinder(boolean supportLayersAndAddOns) {
...
+    }
+
+    private static String toLegacyPathString(String moduleName) {
+        final ModuleIdentifier moduleIdentifier = ModuleIdentifier.fromString(moduleName);
+        final String name = moduleIdentifier.getName();
+        final String slot = moduleIdentifier.getSlot();
+        final StringBuilder builder = new StringBuilder(name.length() + slot.length() + 2);
+        builder.append(name.replace('.', File.separatorChar));
+        builder.append(File.separatorChar).append(slot);
         builder.append(File.separatorChar);
```
ropalka: In this case it doesn't matter if ending slash is present or not but it might be omitted too, doesn't it?

PR #91, opened 2016-02-11
Modules 241 branch 13

ropalka:

PR #118, opened 2016-11-30
MODULES-241 Leaked Alias Modules

ppalaga:
This is a work in progress, please, do not merge yet.


ppalaga:
@dmlloyd and/or @ropalka could you please review?

In https://github.com/jboss-modules/jboss-modules/pull/118/commits/133fa3149b0aaf7b9e426856c27f4e3f1fe06a07#diff-abb90700653f04644b40ca808a8fa8a8R154 , I tried to implement David's idea to handle aliases as "regular modules with a single dependency on the alias target, with the import and export filters set to ALL".

`ModuleClassLoaderAliasReloadTest` comes nearly unchanged from @ropalka 's https://github.com/jboss-modules/jboss-modules/pull/91/commits/b5c5623a7e1867fcbe355f3c431943f615614420#diff-c88ba42fd639b6bd43f1ef4e0bbd1913R37 It is there to verify that MODULES-241 itself is fixed.

I also added some more alias tests to make sure that things work in the same way before and after this change https://github.com/jboss-modules/jboss-modules/pull/118/commits/133fa3149b0aaf7b9e426856c27f4e3f1fe06a07#diff-ca887fb4b8d46c79594bbc0de15d60a4R22

I hope that the deprecation of `AliasModuleSpec` without removinf it is the best thing we can do in this situation https://github.com/jboss-modules/jboss-modules/pull/118/commits/133fa3149b0aaf7b9e426856c27f4e3f1fe06a07#diff-0c5428c3bd022f829208b067d406b38fR30

ppalaga:
> now that AliasModuleSpec is not created we can eliminate all dead code

I was considering it, but TBH I was too lazy to make sure that that code cannot be called from outside JBoss Modules. Because if it can, clearly, we cannot remove it (in a micro release) for compatibility reasons. OK, let me have a look at the call hierarchies and remove if possible.

ppalaga:
The `if (spec instanceof AliasModuleSpec)` condition in `DependencyTreeViewer.print(PrintWriter, String, ModuleSpec, FastCopyHashSet<ModuleIdentifier>, File...)` can be safely removed. It is reachable only by code from within JBoss Modules.

However, `if ( moduleSpec instanceof AliasModuleSpec)` condition in `ModuleLoader.loadModuleLocal(ModuleIdentifier)` cannot be removed safely.

Although the single constructor of `AliasModuleSpec` has only package visibility, it still can be instantiated by third party code using reflection.

Now, how such a third party instance of `AliasModuleSpec` could hit the `if ( moduleSpec instanceof AliasModuleSpec)` condition in `ModuleLoader.loadModuleLocal(ModuleIdentifier)` ?
The `moduleSpec` reference in the condition comes from `findModule(ModuleIdentifier)` where its instantiation is delegated to one of the `ModuleFinder` s. The array of `ModuleFinder` s can be passed through the public constructor. A third party may thus provide its own implementation of `ModuleFinder` that instantiates `AliasModuleSpec` through the reflection. Therefore, we cannot safely exclude the possibility that `moduleSpec` in `ModuleLoader.loadModuleLocal(ModuleIdentifier)` contains an instance of `AliasModuleSpec`.

Does that sound right?

ropalka:
Your analysis is correct. JBoss-Modules users should use **org.jboss.modules.ModuleSpec** API
to create module specifications. Hacking into internals via reflection is not a good practice. But to maintain API compatibility in such cases I agree ModuleLoader should continue accepting AliasModuleSpec too.

ropalka:
BTW such usages of thirdparty module finders creating AliasModuleSpec instances via reflection will lead to MODULES-241 issue.

ppalaga:
ca8a0c6 replaces 133fa31 . In ca8a0c6, the `if (spec instanceof AliasModuleSpec)` block is removed https://github.com/jboss-modules/jboss-modules/pull/118/commits/ca8a0c66240e1cf93cdaf9c67d09ca5b69bc57ba#diff-347492c5b5ad5615dcadb5f2bd3c769fL48 and every unknown `ModuleSpec` throws an exception https://github.com/jboss-modules/jboss-modules/pull/118/commits/ca8a0c66240e1cf93cdaf9c67d09ca5b69bc57ba#diff-347492c5b5ad5615dcadb5f2bd3c769fR56

Is it good to merge like this?

dmlloyd:
Richard's analysis is correct in that we must retain AliasModuleSpec for compatibility.  However there's an aspect that I think was missed: user code may use ``instanceof`` checks to determine if a module spec is or is not an alias.

Thus I think it might be better from a compatibility standpoint to use a slightly different approach: instead of translating an alias ModuleSpec into the equivalent ConcreteModuleSpec when it is built, we should probably keep AliasModuleSpec around and interpret it as the equivalent ConcreteModuleSpec at construction time instead.

WDYT?

ppalaga:
> we should probably keep AliasModuleSpec around and interpret it as the equivalent ConcreteModuleSpec

Sounds safer. Let me try it :)

ppalaga:
c6de7c5 replaces ca8a0c6 . New or noteworthy:

* `AliasModuleSpec` is not deprecated anymore
* `AliasModuleSpec` provides a package visible `asConcreteModuleSpec()` to get the `ConcreteModuleSpec` that is then used in ModuleLoader https://github.com/jboss-modules/jboss-modules/pull/118/commits/c6de7c5483d7331b0a65c226245953aac006db69#diff-0859eb42043997cf133f54a553026232R370

@dmlloyd is this good to merge now?

src/main/java/org/jboss/modules/AliasModuleSpec.java
```diff
@@ -22,21 +22,58 @@
...
     AliasModuleSpec(final ModuleIdentifier moduleIdentifier, final ModuleIdentifier aliasTarget) {
         super(moduleIdentifier);
         this.aliasTarget = aliasTarget;
+
+        DependencySpec aliasTargetDependency = DependencySpec.createModuleDependencySpec(PathFilters.acceptAll(),
+                PathFilters.acceptAll(), null, aliasTarget, false);
+
+        final String mainClass = null;
+        final AssertionSetting assertionSetting = AssertionSetting.INHERIT;
+        final ResourceLoaderSpec[] resourceLoaders = new ResourceLoaderSpec[0];
```
dmlloyd: Small nit, but ideally this would be a constant (would save a couple hundred allocations on server startup).

ppalaga: > ideally this would be a constant (would save a couple hundred allocations on server startup).

I'd like to have your eagle eyes!

We actually have such a constant in `ModuleClassLoader` https://github.com/jboss-modules/jboss-modules/blob/c6de7c5483d7331b0a65c226245953aac006db69/src/main/java/org/jboss/modules/ModuleClassLoader.java#L82

From the performance PoV, using `ModuleClassLoader.NO_RESOURCE_LOADERS` here (call it option A) should be equally OK as declaring a new constant in `AliasModuleSpec` (option B) because both compile to `getstatic`.

However making `AliasModuleSpec` dependent on a random class just because it contains a constant we need does not sound right. Moving `NO_RESOURCE_LOADERS` from `ModuleClassLoader` to a more natural place, such as `ResourceLoaderSpec` (option C) would solve it. 

The question is how paranoid do we want to be w.r.t. compatibility with user code? If moving package visible constants around is forbidden, then option B seems to be the best one. Otherwise, C wins.

WDYT?

dmlloyd: Moving package-private constants is fine.  I think ResourceLoaderSpec is a reasonable place for this.

ppalaga:
73dcac8 replaces c6de7c5 . `ModuleClassLoader.NO_RESOURCE_LOADERS` constant was moved to `ResourceLoaderSpec` and `AliasModuleSpec` uses it from there.

Given that https://issues.jboss.org/browse/MODULES-241 requires a fix in all of 1.1 through 1.5 and master, should I send separate PRs to all those branches?

dmlloyd:
Looks great.  Yeah if you could send fixes for 1.1 through 1.5 *and* 1.x that would save me some work.  Thanks!

ppalaga:
@dmlloyd while the PRs for 1.1 https://github.com/jboss-modules/jboss-modules/pull/120 and 1.2 https://github.com/jboss-modules/jboss-modules/pull/121 are fully analogic to the present one for 1.3, the in the ones for the newer branches https://github.com/jboss-modules/jboss-modules/pull/122 https://github.com/jboss-modules/jboss-modules/pull/123 https://github.com/jboss-modules/jboss-modules/pull/119 and https://github.com/jboss-modules/jboss-modules/pull/124 I had to adapt the `ModuleClassLoaderAliasTest`: 

compare https://github.com/ppalaga/jboss-modules/blob/73dcac8a45e8c11fa06296d15dd5fb2f6a03a5a5/src/test/java/org/jboss/modules/ModuleClassLoaderAliasTest.java#L48-L52 vs https://github.com/ppalaga/jboss-modules/blob/06afe96ccf588c2042a529c9cc2a1ca0e3eccca0/src/test/java/org/jboss/modules/ModuleClassLoaderAliasTest.java#L48-L52

and https://github.com/ppalaga/jboss-modules/blob/73dcac8a45e8c11fa06296d15dd5fb2f6a03a5a5/src/test/java/org/jboss/modules/ModuleClassLoaderAliasTest.java#L87-L91 vs https://github.com/ppalaga/jboss-modules/blob/06afe96ccf588c2042a529c9cc2a1ca0e3eccca0/src/test/java/org/jboss/modules/ModuleClassLoaderAliasTest.java#L87-L91

I assume the difference in the behavior is known to you. Anyway, for each of the named branches, the tested behavior is the same before and after the MODULES-241 fix.

PR #127, opened 2016-12-20
[MODULES-277] Master fix

ropalka:


dmlloyd:
Replaced with generalized Java 9 support.

PR #133, opened 2017-04-21
Fixing JBoss module aliases issues

ropalka:
https://issues.jboss.org/browse/JBEAP-10083
https://issues.jboss.org/browse/MODULES-282
https://issues.jboss.org/browse/MODULES-241


src/main/java/org/jboss/modules/ModuleLoader.java
```diff
@@ -493,13 +493,31 @@ protected final Module loadModuleLocal(String name) throws ModuleLoadException {
...
+                    if (aliases == null) continue; // race condition detected, retry
+                    synchronized (aliases) {
+                        if (aliases == aliasedModule.aliases) {
+                            // no race-condition with removal, go ahead
+                            aliases.add(moduleSpec.getName());
+                            newFuture.setModule(module = aliasedModule);
+                            ok = true;
+                            break;
+                        }
+                    }
```
ppalaga: It would be handy to have a trace statement somewhere in this if `AliasModuleSpec` branch too, perhaps just before the `break` above. Something like `log.trace("Added module %s as alias of %s from %s", moduleSpec.getName(), aliasModuleSpec.getAliasName(), this);`

src/main/java/org/jboss/modules/ModuleLoader.java
```diff
@@ -565,6 +583,21 @@ protected final void unloadModuleLocal(Module module) throws SecurityException {
         final String id = module.getIdentifier().toString();
         final FutureModule futureModule = moduleMap.get(id);
         if (futureModule != null && futureModule.module == module) {
+            final Set<String> aliases = module.aliases;
+            if (aliases != null) {
+                synchronized (aliases) {
+                    if (aliases.contains(id)) {
```
ppalaga: I asked this once already, citing me and you and commenting inline:

> >  if (!module.aliases.contains(id)) condition necessary?
>
> The idea is alias module can be removed without invalidating aliased module.
> (In other words there is possibility alias module is unloaded at runtime).

I do not think unloading an alias module is possible. 

Let me explain it on an example: Let's say module `M1A` is an alias of module `M1` and there are no other modules in our universe. 

There will be two entries `M1A` and `M1` in `moduleMap`, but there will be just one instance of `org.jboss.modules.Module` in the JVM. The `name` of that `org.jboss.modules.Module` instance will be `M1`. Because there will be no `org.jboss.modules.Module` instance with name `M1A`, it is simply not be possible to call `unloadModuleLocal()` for that instance. 

The only possible call of `unloadModuleLocal()` is with `org.jboss.modules.Module` having name `M1` therefore `if (aliases.contains(id))` cannot be true at any reasonable circumstances.

That said, I think the whole `if (aliases.contains(id)) ... else ...` statment should be replaced with the content of the `else` branch.

> On the other hand when non alias module is unloaded, all its aliases should be invalidated too.
>
> >  Will module.aliases ever contain the id? Does it make any sense to register an alias for a module that is the same as its id?
>
>I don't think so.

ropalka:
Great job @ppalaga ! You identified anoher bug. Unloading of alias module shouldn't result into unload of aliased module. The if statement in question is really necessary. I updated the PR.

src/main/java/org/jboss/modules/ModuleLoader.java
```diff
@@ -574,13 +575,33 @@ protected final Module findLoadedModuleLocal(String name) {
      * @param module the module to unload
      * @throws SecurityException if an attempt is made to unload a module which does not belong to this module loader
      * @throws SecurityException if the module was not defined by this module loader
+     * @deprecated use {@link #unloadModuleLocal(String,Module)} instead
      */
+    @Deprecated
     protected final void unloadModuleLocal(Module module) throws SecurityException {
+        unloadModuleLocal(null, module);
```
ppalaga: For the sake of cleanness of the contract and maintainability, I vote for making the new `unloadModuleLocal()` to accept the module name only rather than name and module. Inside the proper `unloadModuleLocal()` we lookup in the the `moduleMap` anyway.

src/main/java/org/jboss/modules/ModuleLoader.java
```diff
@@ -574,13 +575,33 @@ protected final Module findLoadedModuleLocal(String name) {
...
+    protected final void unloadModuleLocal(final String moduleName, final Module module) throws SecurityException {
         final ModuleLoader moduleLoader = module.getModuleLoader();
         if (moduleLoader != this) {
             throw new SecurityException("Attempted to unload " + module + " from a different module loader");
         }
-        final String id = module.getIdentifier().toString();
+        final String id = moduleName == null ? module.getIdentifier().toString() : moduleName;
         final FutureModule futureModule = moduleMap.get(id);
         if (futureModule != null && futureModule.module == module) {
             final Set<String> aliases = module.aliases;
```
ppalaga: I vote for simplifying of 

```
                    if (aliases.contains(id)) {
                         // removing alias module
                         aliases.remove(id);
                     } else {
                         // cleaning aliased module and all its aliases
                         module.aliases = null;
                         for (String alias : aliases) {
                             moduleMap.remove(alias);
                         }
                     }
```

to

```
                    if (!aliases.remove(id)) {
                        /* if the id is an alias, just remove it from the set of aliases
                         * othewise clean the aliased module and all its aliases */
                        module.aliases = null;
                        moduleMap.remove(id);
                        for (String alias : aliases) {
                            moduleMap.remove(alias);
                        }
                    }
```

In this way, we are avoiding the `aliases.contains(id)` call and the resulting state is the same.

src/main/java/org/jboss/modules/ModuleLoader.java
dmlloyd: The contract of the two-argument version seems useful: "Unload the module name xxx if it still refers to actual module yyy".  This might be useful to avoid problems in racy situations (like concurrent deploy and undeploy), but such a method should also return a boolean to indicate that the change was successfully made.

ropalka:
First proposal would make sense in sequential programming.
But in concurrent usecases it is the right thing to do, because
unloadModuleLocal() is not atomic operation and can be executed
concurrently by multiple threads. Thus we're passing Module
reference to ensure the correct 'snapshoted' module reference is
removed. This approach is similar to ConcurrentHashMap.remove(Object,Object) method.

Second proposal is good one, I've updated latest commit to incorporate it.

ropalka:
I've incorporated David's proposal for boolean return value too.

src/main/java/org/jboss/modules/ModuleLoader.java
ppalaga: OK, I see the reason to keep unloadModuleLocal(String, Module) with two args. 

However, if the caller is supposed to know which name-module association should be removed, why the devious step to `null` ? I vote for requiring a non-null `moduleName` in unloadModuleLocal(String, Module). In that way, we can avoid the call to the deprecated `Module.getIdentifier()` and make the execution path more straightforward.

src/main/java/org/jboss/modules/ModuleLoader.java
```diff
@@ -574,21 +575,41 @@ protected final Module findLoadedModuleLocal(String name) {
...
+
+    /**
+     * Unload a module from this module loader.  Note that this has no effect on existing modules which refer to the
+     * module being unloaded.  Also, only modules from the current module loader can be unloaded.  Unloading the same
+     * module more than once has no additional effect.  This method only removes the mapping for the module; any running
+     * threads which are currently accessing or linked to the module will continue to function, however attempts to load
+     * this module will fail until a new module is loaded with the same name.  Once this happens, if all references to
+     * the previous module are not cleared, the same module may be loaded more than once, causing possible class duplication
+     * and class cast exceptions if proper care is not taken.
+     *
```
ppalaga: I vote for adding here something like

> If the given {@code moduleName} is an alias of the given {@code module}, then only the association of the alias  with the {@code module} is removed and the association of the {@code module} with its canonical name and other aliases is left intact.
>
> If the given {@code moduleName} is the canonical name of the given {@code module}, then all alias associations are removed together with the canonical {@code moduleName} association.

src/main/java/org/jboss/modules/ModuleLoader.java
```diff
@@ -556,17 +575,53 @@ protected final Module findLoadedModuleLocal(String name) {
...
         if (futureModule != null && futureModule.module == module) {
-            moduleMap.remove(id, futureModule);
+            final Set<String> aliases = module.aliases;
+            if (aliases != null) {
+                synchronized (aliases) {
+                    if (aliases.remove(id)) {
+                        return true;
+                    } else {
+                        // cleaning aliased module and all its aliases
+                        module.aliases = null;
```
ppalaga: If we return `true` here, we have not performed `moduleMap.remove(id, futureModule)` which we should.

ppalaga: Actually, now that we have the `moduleName` method argument, it is perhaps cheaper to use `if (id.equals(module.getName()))` to figure out whether the `id` is an alias or a canonical name. 

```
            if (aliases != null) {
                synchronized (aliases) {
                    if (!id.equals(module.getName())) {
                        // id is an alias
                        aliases.remove(id);
                    } else {
                        // id is a canonical name
                        // clean all aliases
                        module.aliases = null;
                        for (String alias : aliases) {
                            moduleMap.remove(alias);
                        }
                    }
                }
            }
            return moduleMap.remove(id, futureModule);
```

ropalka:
I stashed last two commits into one @ppalaga & @dmlloyd .
I incorporated almost whole aforementioned valuable feedback from Peter
except the last one because in most cases aliases modules are such a rare
that 99% of modules are regular modules with zero aliases. So I think
aliases.remove() on empty set should be more cheaper operation than String comparison.

ropalka:
Thanks @ppalaga I merged and squashed your commits into one.
I prefer to keep your contribution separate because it was significant one
and should be tracked in modules history as yours. 
How about you @dmlloyd any comments and / or objections on this PR?

ropalka:
I know you're very busy @dmlloyd lately but anyway:
Could you please review and either approve or decline this PR?

dmlloyd:
It looks pretty good.  At this point I think we can fix problems as they arise.

PR #136, opened 2017-05-15
WIP - second try of jarfs

ctomc:


rsvoboda:
Running some checks on patched server (latest DR build + jboss-modules from Tomaz's branch)

Scope:
 * startup times
 * memory footprint
 * deployment performance - big ear file 59 MB
 * deployment performance - many jars (2000+) in WEB-INF/lib/ of the war deployment which is ~4MB

Details for last case
Using script from https://raw.githubusercontent.com/sdaschner/multi-module-performance/master/multi-hierarchical/create.sh with small adjustment
```
data=(a b c d e f g h i j k l m n o p q r s t u v w x y z)
```
replaced with 
```
data=(`for x in {a..z}; do for y in {a..z}; do echo "${x}_${y}"; done; done`)
```
to have really a lot of modules in maven project
```

**Comment by rsvoboda** [comment]
I have bad news, I see worse results for all scenarios I described.

I will share them with @ctomc via  HipChat

**Comment by ctomc** [comment]
yeah, we get less locking but other aspects are not as good. 
will do some more changes to see what can be improved.

**Comment by rsvoboda** [comment]
ping me once you have something for next round of testing

File: src/main/java/org/jboss/modules/ResourceLoaders.java [hunk 1]
```diff
@@ -75,10 +77,21 @@ public static IterableResourceLoader createIterableFileResourceLoader(final Stri
      * @param jarFile the backing JAR file
      * @return the resource loader
      */
+    public static ResourceLoader createJarResourceLoader(final String name, final Path jarFile) throws IOException {
+        return new JarResourceLoader(name, jarFile, AccessController.getContext());
+    }
+
+    @Deprecated
     public static ResourceLoader createJarResourceLoader(final String name, final JarFile jarFile) {
-        return new JarFileResourceLoader(name, jarFile);
```
rsvoboda: @ctomc - is possible to do some reasonable performance comparison of  old JarFileResourceLoader vs. new JarResourceLoader? Thinking what would do the difference I noticed and the loader class is my tip...

ctomc: sure, we need to do that before moving to this. 
From my tests new approach works much better when you have lots of deployments and there is lots of concurrency going on.

I also know what is the reason for slow down you see, am looking into finding best solution for it.

rsvoboda:
Rebased branch still suffers with slower boot time, increased memory footprint, and longer time for deployment - especially visible in hierarchical case.

PR #140, opened 2017-07-10
[MODULES-294] Minor boot time improvements

fossnova:
This is the maximum (for now) I was able to accelerate WildFly boot time with these accumulated jboss-modules fixes . I benchmarked the changes (see mail for reference) and I can see almost 2 % boot time improvement. Let me know if there are some changes you don't like I can exclude them from this pull request. I hoped for better boot time improvements. Maybe next time.

### DIFF
--- a/src/main/java/org/jboss/modules/JarFileResourceLoader.java
+++ b/src/main/java/org/jboss/modules/JarFileResourceLoader.java
@@ -141,19 +141,9 @@ final class JarFileResourceLoader extends AbstractResourceLoader implements Iter
                 spec.setCodeSource(codeSource);
                 return spec;
             } else if (size <= (long) Integer.MAX_VALUE) {
-                final int castSize = (int) size;
-                byte[] bytes = new byte[castSize];
-                int a = 0, res;
-                while ((res = is.read(bytes, a, castSize - a)) > 0) {
-                    a += res;
-                }
-                // consume remainder so that cert check doesn't fail in case of wonky JARs
-                while (is.read() != -1) {
-                    //
-                }
-                // done
+                // size known
+                spec.setBytes(getClassBytes(is, (int) size));
                 CodeSource codeSource = createCodeSource(entry);
-                spec.setBytes(bytes);
                 spec.setCodeSource(codeSource);
                 return spec;
             } else {
@@ -172,6 +162,20 @@ final class JarFileResourceLoader extends AbstractResourceLoader implements Iter
         }
         baos.close();
         return baos.toByteArray();
+    }
+
+    private byte[] getClassBytes(final InputStream is, final int castSize) throws IOException {
+        assert holdsLock(this);
+        final byte[] bytes = new byte[castSize];
+        int a = 0, res;
+        while ((res = is.read(bytes, a, castSize - a)) > 0) {
+            a += res;
+        }
+        // consume remainder so that cert check doesn't fail in case of wonky JARs
+        while (is.read() != -1) {
+            //
+        }
+        return bytes;
     }
 
     // this MUST only be called after the input stream is fully read (see MODULES-201)
--- a/src/main/java/org/jboss/modules/JarFileResourceLoader.java
+++ b/src/main/java/org/jboss/modules/JarFileResourceLoader.java
@@ -131,52 +131,52 @@ final class JarFileResourceLoader extends AbstractResourceLoader implements Iter
             // no such entry
             return null;
         }
-            final ClassSpec spec = new ClassSpec();
-            final long size = entry.getSize();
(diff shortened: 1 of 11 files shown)


src/main/java/org/jboss/modules/JarFileResourceLoader.java
```diff
@@ -131,52 +131,52 @@ public synchronized ClassSpec getClassSpec(final String fileName) throws IOExcep
             // no such entry
             return null;
         }
-            final ClassSpec spec = new ClassSpec();
```
dmlloyd: Doing this as a separate step is not necessary since we can always do a diff with the `-w` flag, or (in GitHub) add `?w=1` to the query parameters, to view the diff without whitespace changes.

src/main/java/org/jboss/modules/JarFileResourceLoader.java
```diff
@@ -137,15 +137,14 @@ public synchronized ClassSpec getClassSpec(final String fileName) throws IOExcep
             // size unknown
             spec.setBytes(getClassBytes(entry));
             spec.setCodeSource(createCodeSource(entry));
-            return spec;
```
dmlloyd: These kinds of simplifications are highly subjective/opinion-based.  In general it's a better practice to leave code as is when there is an established style (even if you disagree with it) unless there is some objective reason (like improved performance).  One danger is that you may get into a pointless "commit war" with someone else whose opinions are just as strong (but opposing).

src/main/java/org/jboss/modules/JarFileResourceLoader.java
```diff
@@ -61,7 +61,7 @@
  * @author <a href="mailto:ropalka@redhat.com">Richard Opalka</a>
  */
 final class JarFileResourceLoader extends AbstractResourceLoader implements IterableResourceLoader {
-    private static final String INDEX_FILE = "META-INF/PATHS.LIST";
+    private static final String INDEX_FILE = "META-INF/INDEX.LIST";
```
dmlloyd: I believe this would be an incompatible change.

src/main/java/org/jboss/modules/JarFileResourceLoader.java
```diff
@@ -217,34 +219,12 @@ public String getLibrary(final String name) {
...
-                path = PathUtils.canonicalize(path);
-                if (File.separatorChar != '/') {
-                    // optimizes away on platforms with /
-                    path = path.replace(File.separatorChar, '/');
-                }
-                if (PathUtils.isRelative(path)) {
-                    // should not be possible, but the JDK thinks this might happen sometimes..?
-                    path = "/" + path;
-                }
-                if (path.startsWith("//")) {
```
dmlloyd: This was necessary for Windows.  Is this logic replicated anywhere?

src/main/java/org/jboss/modules/JarFileResourceLoader.java
```diff
@@ -68,7 +67,7 @@
     private final URL rootUrl;
     private final String relativePath;
     private final File fileOfJar;
-    private final URI jarURI;
+    private final String uriPrefix;
     private volatile List<String> directory;
```
dmlloyd: Do these changes make a concrete performance difference?  If no, let's drop them.

src/main/java/org/jboss/modules/filter/GlobPathFilter.java
```diff
@@ -26,12 +26,14 @@
...
  * @author John E. Bailey
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
+ * @author <a href="mailto:ropalka@redhat.com">Richard Opalka</a>
  */
 final class GlobPathFilter implements PathFilter {
     private static final Pattern GLOB_PATTERN = Pattern.compile("(\\*\\*?)|(\\?)|(\\\\.)|(/+)|([^*?]+)");
 
     private final String glob;
     private final Pattern pattern;
+    private final int hashCode;
```
dmlloyd: Did these changes to the filter make a performance difference?

src/main/java/org/jboss/modules/ModuleLoader.java
```diff
@@ -859,15 +859,15 @@ void addClassLoadTime(final long time) {
     }
 
     void incScanCount() {
-        if (Metrics.ENABLED) scanCount.getAndIncrement();
```
dmlloyd: Why move the check out of the method?  It should be inlined anyway.

ropalka:
You confirmed my reservations @dmlloyd I had to this pull request.
I'm closing this pull request because I agree its value is minimal (if not zero).

PR #146, opened 2017-12-08
MODULES-319 Add module.xml to select module contents at run time

stuartwdouglas:


src/main/java/org/jboss/modules/xml/SystemPropertyConditionBuilder.java
```diff
@@ -0,0 +1,47 @@
...
+        private final boolean equal;
+
+        private Condition(String name, String value, boolean equal) {
+            this.name = name;
+            this.value = value;
+            this.equal = equal;
+        }
+
+        public boolean get() {
+            String pval = System.getProperty(name);
```
ropalka: This call should be privileged.

src/main/java/org/jboss/modules/xml/ModuleXmlParser.java
```diff
@@ -902,20 +911,31 @@ private static void parseNativeArtifact(final MavenResolver mavenResolver, final
...
+        final SystemPropertyConditionBuilder conditionBuilder = new SystemPropertyConditionBuilder();
+        final Set<String> encountered = new HashSet<>();
         int eventType;
         for (;;) {
             eventType = reader.nextTag();
             switch (eventType) {
                 case END_TAG: {
                     try {
-                        createMavenNativeArtifactLoader(mavenResolver, name, reader, specBuilder);
+                        if(conditionBuilder.resolve()) {
```
ropalka: Not a problem @stuartwdouglas but I noticed you're using strange indentation for conditionals (missing space between conditional type and first left bracket). Would be nice to have it synced with others.

src/main/java/org/jboss/modules/xml/SystemPropertyConditionBuilder.java
dmlloyd: This does not actually need to be privileged; rather, the entire call to parse the module.xml should be privileged.  At any rate we don't want to use a lambda in there.

stuartwdouglas: I have updated it to do the parsing in priviledged blocks

stuartwdouglas:
@dmlloyd do you want any further changes for this?

src/main/java/org/jboss/modules/xml/SystemPropertyConditionBuilder.java
dmlloyd: What I meant to say is, the call to parse the module.xml should *already* be privileged, meaning we don't need doPrivileged at all.

dmlloyd:
Just added one comment to remove the doPrivileged thing; other than that it looks OK.

src/main/java/org/jboss/modules/ParseModuleXmlPriviledgedAction.java
```diff
@@ -0,0 +1,30 @@
+package org.jboss.modules;
+
+import java.io.InputStream;
+import java.security.PrivilegedExceptionAction;
+
+import org.jboss.modules.xml.ModuleXmlParser;
+
+class ParseModuleXmlPriviledgedAction implements PrivilegedExceptionAction<ModuleSpec> {
```
dmlloyd: This class now seems to be unused.

src/main/java/org/jboss/modules/xml/ModuleXmlParser.java
```diff
@@ -1030,6 +1059,56 @@ private static void parseResourceRoot(final ResourceRootFactory factory, final S
         }
     }
 
+    private static void parseConditions(XmlPullParser reader, SystemPropertyConditionBuilder conditionBuilder) throws XmlPullParserException, IOException {
+        atLeast1_7(reader);
```
dmlloyd: The return value is unused here.

src/main/java/org/jboss/modules/xml/ModuleXmlParser.java
```diff
@@ -902,20 +911,31 @@ private static void parseNativeArtifact(final MavenResolver mavenResolver, final
...
+        final SystemPropertyConditionBuilder conditionBuilder = new SystemPropertyConditionBuilder();
+        final Set<String> encountered = new HashSet<>();
         int eventType;
         for (;;) {
             eventType = reader.nextTag();
             switch (eventType) {
                 case END_TAG: {
                     try {
-                        createMavenNativeArtifactLoader(mavenResolver, name, reader, specBuilder);
+                        if (conditionBuilder.resolve()) {
```
dmlloyd: Just out of curiousity, why defer the resolution of the condition builder to the end tag, instead of just parsing it to a boolean right away?

stuartwdouglas: No real reason, I don't think it makes much difference either way.

PR #147, opened 2018-01-18
[MODULES-323] Fixing NPEs

ropalka:


dmlloyd:
Should a similar change be done to __XMLInputFactory, __XMLReaderFactor, etc.?

ropalka:
You're right of course @dmlloyd my fault.

PR #152, opened 2018-02-13
changed PolicyExpander to be public

jasinner:


jasinner:
Will resubmit with correct branch name MODULES-331

PR #153, opened 2018-02-13
made PolicyExpander public

jasinner:


jasinner:
If we're going to do another 1.x release can we also backport this to the 1.x branch?

ropalka:
I can't open MODULES-336 issue (seems it doesn't exist). What is your usecase you are trying to achieve? Generally speaking exposing classes to public needs reasonable consideration @jasinner .

ropalka:
Found it, it is MODULES-331 not MODULES-336.

ropalka:
I think we can expose this utility class to public @dmlloyd .
But class should be both **public** and **final** @jasinner .
And should introduce private NOOP constructor to forbid instantiations.

dmlloyd:
Is there a reason that the wildfly-common property expander cannot be used?

dmlloyd:
Assuming there's a good reason to use this code instead of the WFCOM code, it needs:

* A JIRA # in the commit and PR title
* The class should be public/final with private ctor as Richard says
* The class should be moved to the org.jboss.modules.security package

dmlloyd:
Also if this is to be merged into two branches, I would prefer it if the commit were based directly on a common ancestor commit (preferably the one that introduced the class in the first place).

jasinner:
See https://issues.jboss.org/browse/MODULES-331, won't implement using PolicyExpander in JBoss Modules.

PR #173, opened 2018-03-28
[MODULES-254] Support Java 9 module dependencies natively

dmlloyd:


src/main/java/org/jboss/modules/JDKModuleFinder.java
```diff
@@ -0,0 +1,242 @@
...
+            "java.management",
+            "java.management.rmi",
+            "java.naming",
+            "java.prefs",
+            "java.rmi",
+            "java.scripting",
+            "java.security.jgss",
+            "java.security.sasl",
+            "java.sql",
+            "java.sql.rowset",
```
ropalka: I see you ommited java.transaction.xa - is it intentional?

dmlloyd: Yes.  The set is meant to mimic the set of Java 9 modules, minus the ``java.se.ee`` ones.  The ``java.transaction.xa`` module does not exist in Java 9 or 10.

PR #196, opened 2018-05-01
[MODULES-365] Using URLStreamHandler explicitly in the URL constructor

gastaldi:
This is necessary for FURNACE-153


dmlloyd:
Merged as #198 and #199.  Thanks!

PR #197, opened 2018-05-01
[MODULES-366] ModuleLoader.installMBeanServer is now public

gastaldi:
Some frameworks need it:
- JBoss Forge: https://github.com/forge/furnace/blob/fc9c2dcd3c8706ae1f587c6c347b6b4387faead9/container/src/main/java/org/jboss/forge/furnace/impl/modules/AddonModuleLoader.java#L326-L338
- WildFly Swarm: https://github.com/wildfly-swarm/wildfly-swarm/blob/d16c244a4f3877396f0ff190f4fc355e14dcf501/core/container/src/main/java/org/wildfly/swarm/Swarm.java#L854-L862


dmlloyd:
Fixed by #265.

PR #206, opened 2018-05-09
[MODULES-369] Add all known non-exported packages to Java 8 emulation modules

dmlloyd:


src/main/resources/jdk-module-paths/java.base
```diff
@@ -49,7 +49,86 @@ javax/security/auth/login
 javax/security/auth/spi
 javax/security/auth/x500
 javax/security/cert
-# These are Java 8 only, but required for correct operation
+# Non-exported
+com/sun/java/util/jar/pack
+com/sun/net/ssl
+com/sun/net/ssl/internal/www/protocol/https
```
darranl: At this point we are also missing 'com/sun/net/ssl/internal'

PR #218, opened 2018-09-07
[MODULES-375] Check if the artifact was successfully resolved before …

jamezp:
…attempting to create the loader.
https://issues.jboss.org/browse/MODULES-375

There doesn't seem to be a need for an upstream fix and the `ResourceLoader` is created in the `MavenArtifactUtil` which already checks for a `null` resolved artifact.


Ladicek:
This is very much needed in the main branch as well. `MavenArtifactUtil` checking for a `null` artifact doesn't really matter, as it's the `file.getPath()` call that throws the NPE. I'll submit a PR.

Ladicek:
Submitted #234.

PR #221, opened 2018-10-15
[MODULES-379] Ability to attach properties to module dependencies

aloubyansky:
https://issues.jboss.org/browse/MODULES-379

This commit adds:

- PropertyBuilder interface with addProperty(name, value) method implemented by builder classes for ModuleSpec and ModuleDependencySpec;
- module-1_9.xsd which adds properties to module/dependencies/module.

NOTE: I moved setting assertion settings from the parser to ModuleSpec builder. Please let me know if that's a bad idea.

Thanks!

### DIFF
--- a/src/main/java/org/jboss/modules/xml/ModuleXmlParser.java
+++ b/src/main/java/org/jboss/modules/xml/ModuleXmlParser.java
@@ -108,6 +108,13 @@ public final class ModuleXmlParser {
         }
     }
 
+    /**
+     * XML parsing callback for property elements.
+     */
+    private interface PropertiesCallback {
+        void parsedProperty(String name, String value);
+    }
+
     private ModuleXmlParser() {
     }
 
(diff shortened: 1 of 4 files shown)


dmlloyd:
Is there a reason that you needed the properties to be available via the DependencySpec API?  Are you using JBoss Modules to parse these in Galleon?

aloubyansky:
Unfortunately I am not using the parser from JBoss Modules itself. I don't see how I could do that. Would be great though. So, no, I won't actually use ModuleDependencySpec API.

dmlloyd:
In that case, I think it would be better to simply modify the parser to skip & ignore nested `properties` elements rather than parsing them into the dependency spec model.  If we need the info at run time, we can add that part of the patch later.

aloubyansky:
Sure, no problem. I've adjusted the commit.

PR #233, opened 2019-01-29
[MODULES-372] provide our own "java.se" module on JDK 9+

Ladicek:
The `java.se` module in JDK 9+ isn't "available" by default. However,
it's merely an aggregator of other modules that _are_ "available".
This commit provides our own `java.se` module even on JDK 9+, instead
of delegating to JPMS. The new `java.se` module aggregates the same
modules as the `java.se` module in the JDK, the only difference is
that it's constructed on the JBoss Modules level and hence doesn't
require users to start JVM with `--add-modules java.se`.


Ladicek:
Also if we could get a release with this (1.9? 1.10?), that would be great for us (Thorntail).

Ladicek:
Rebased.

gastaldi (approved):
LGTM and we could use a release with this patch in Forge too

PR #243, opened 2019-09-20
[MODULES-393] Implement a proxy for any log manager which includes a ModuleFinder and a ModuleFinderFactory

jamezp:
https://issues.jboss.org/browse/MODULES-393
https://issues.jboss.org/browse/MODULES-394

This relates to [LOGMGR-261](https://issues.jboss.org/browse/LOGMGR-261) and [WFCORE-4674](https://issues.jboss.org/browse/WFCORE-4674) and having the JBoss Log Manager on the boot class path. This is a bit of a hack to use a proxy and implement an interface that [LOGMGR-261](https://issues.jboss.org/browse/LOGMGR-261) will introduce. For better or worse, the reviewer can decide :), I've implemented this in a way where it could be used with any log manager. Not that there likely will be others, but it was easy enough to do so I did it.

The MODULES-394 fix is just a test fix for a failing test when running against Java 8.

### DIFF
--- a/src/main/java/org/jboss/modules/Main.java
+++ b/src/main/java/org/jboss/modules/Main.java
@@ -27,7 +27,10 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.PrintWriter;
 import java.lang.management.ManagementFactory;
+import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -39,6 +42,7 @@ import java.security.ProtectionDomain;
 import java.security.Provider;
 import java.security.Security;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


src/main/java/org/jboss/modules/Main.java
```diff
@@ -520,11 +524,16 @@ public static void main(String[] args) throws Throwable {
...
+            // Special handling is required for cases when a log manager is on the boot class path
+            configureLogManagerModuleFinder(logManagerName);
             if (LogManager.getLogManager().getClass() == LogManager.class) {
                 System.err.println("WARNING: Failed to load the specified log manager class " + logManagerName);
             } else {
                 Module.setModuleLogger(new JDKModuleLogger());
             }
+        } else {
+            // Special handling is required for cases when a log manager is on the boot class path
+            configureLogManagerModuleFinder(System.getProperty("java.util.logging.manager"));
```
jamezp: I'm guessing this is an unlikely scenario and maybe we shouldn't do this, but I've kept it for now. Let me know if we should remove this part.

jamezp:
I forgot to ask what other branches this might need to be ported to. I assume master, but does 1.x need this change too? Maybe 2.x as well.

src/main/java/org/jboss/modules/Main.java
dmlloyd: TBH I'd go way simpler, and avoid reflection (which can come back to bite later).  Just add a block of code in a method like this:

```java
try {
    LogModuleResolverHolder.setResolver(...);
} catch (LinkageError ignored) {
}
```

Then make sure JBoss Modules is compiled against LogManager as a provided, optional dep.  Since JBoss LogManager depends on JBoss Modules, you could break the cycle by putting the couple of classes in a tiny third project, or by using the dependency plugin to fetch it in and add it to compilation as an `additionalClassPathElement`, or just add a `stubs` where a stubbed-out version of that class is included instead of depending on LogManager (you need a separate compiler plugin execution to compile the stubs and again use `additionalClassPathElement` to make sure they're visible to the compilation phase of Modules).

OK not really simpler from a Maven POV but from a code POV it's, like, a couple lines only!

jamezp: A small third project may make sense. Just a simple `ModuleFinder` project which any project could use as this is likely a common pattern for Java 8/Java 11 projects that need something special.

That said that's another JAR that will be required on the boot class path and the system packages.

ropalka: I guess https://github.com/jboss-modules/jboss-modules/pull/244 is a follow up on this discussion, correct guys?

jamezp: @ropalka yes. I don't think there has been a determination on how we want this done. So far this is the only way I've got it to work.

PR #244, opened 2019-09-25
Inject module resolver into bootstrap class path

dmlloyd:


ropalka (approved):
This proposal LGTM

dmlloyd:
We didn't end up going with this approach, so I'm closing this.  Agent mode was introduced instead.

PR #245, opened 2019-10-07
Logger finder

jamezp:
@dmlloyd This is an idea of the logger finder we talked about a couple weeks ago. WDYT?


ropalka (commented):
LGTM

jamezp:
Given the change in #249 I no longer think we'll need this.

PR #246, opened 2019-10-29
[MODULES-395] Initiallize to AllPermissions the default permission collection for the ModuleSpec builder

yersan:
I'll backport to master if finally the patch is accepted.

Jira Issue: https://issues.jboss.org/browse/MODULES-395

### DIFF
--- a/src/main/java/org/jboss/modules/ModuleSpec.java
+++ b/src/main/java/org/jboss/modules/ModuleSpec.java
@@ -21,6 +21,7 @@ package org.jboss.modules;
 import java.lang.instrument.ClassFileTransformer;
 import java.security.AllPermission;
 import java.security.PermissionCollection;
+import java.security.Permissions;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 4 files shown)


dmlloyd:
It seems like the CI failures might be related?

yersan:
Looking ...

yersan:
I got the same errors on 1.9 branch, without this patch, so it is unrelated. It looks like a problem downloading the xerces jar file, seems to be empty. I'm looking for the root error. The same test fails with master branch too.

dmlloyd:
I reran the upstream tests and it looks like it's broken there too.  The tests have to be fixed before I can merge anything else though.

yersan:
@dmlloyd Added a second commit to fix the tests. The nexus maven URL was giving us a 302 when the files are being downloaded by http. So, changed the URL to https and add code to follow up links if a 302 is received again.

dmlloyd:
Great, thanks!

yersan:
@dmlloyd I was going to backport the change but I'm not sure about how the branching model for JBoss modules is, do you need to backport the change to the other branches master, 1.x and 2.x?

dmlloyd:
I can merge it into the other branches.  I've been using a strategy of merging the downstream branches into the upstream ones periodically.

Thanks!

PR #247, opened 2019-10-31
Fix compilation error on JDK 13

tobiasstadler:
In JDK 13 `newFileSystem(java.nio.file.Path, java.util.Map)` was added to `java.nio.file.FileSystems`. Fix compilation by casting `null` to `ClassLoader` so the compiler calls `newFileSystem(java.nio.file.Path, java.lang.ClassLoader)`


dmlloyd:
We require that all contributions be made under the terms of the MIT License, http://www.opensource.org/licenses/mit-license.php. Do you agree to the these terms?

tobiasstadler:
Yes, I do

dmlloyd:
Thanks!

PR #249, opened 2019-11-12
[MODULES-393] Introductory agent mode

dmlloyd:
Introduce a simple agent mode for JBoss Modules.

- Enable agent mode by adding `-javaagent:jboss-modules.jar` to the JVM's arguments
- Add agents to the run time by adding `-javaagent:your-agent.jar` to the arguments of JBoss Modules
- Agent arguments are not supported yet
- It only works if JBoss Modules was added as an agent per the first step, but in the future a self-attach mode could be supported as well

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -222,6 +222,10 @@
                             <Multi-Release>true</Multi-Release>
                             <Jar-Version>${project.version}</Jar-Version>
                             <Jar-Name>${project.artifactId}</Jar-Name>
+                            <Agent-Class>org.jboss.modules.ModularAgent</Agent-Class>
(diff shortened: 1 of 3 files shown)


jamezp (approved):

dmlloyd:
Merged.  Self-attach is really nontrivial and I don't think it's worth the effort at the moment.

jamezp:
I agree @dmlloyd. TBH if we really want we might be able to do that in the scripts. It will be quite easy in the launcher API and I may do it there.

dmlloyd:
:+1: that's what swayed my opinion.

PR #252, opened 2020-01-30
[MODULES-378] Module resource loader follows symbolic links (1.8.x)

spyrkob:
1.8 port of #250

JIRA: https://issues.redhat.com/browse/JBEAP-18544

Upstream issue: https://issues.redhat.com/browse/JBEAP-18585 / https://issues.redhat.com/browse/MODULES-378
Upstream PR: #251 / #250

### DIFF
--- a/src/main/java/org/jboss/modules/PathResourceLoader.java
+++ b/src/main/java/org/jboss/modules/PathResourceLoader.java
@@ -24,6 +24,7 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
 import java.nio.file.FileSystems;
+import java.nio.file.FileVisitOption;
 import java.nio.file.Files;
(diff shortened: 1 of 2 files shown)


ropalka:
Why is backport to 1.8 branch needed @spyrkob ?

spyrkob:
@ropalka we need to fix the issue on EAP 7.2.x stream which is using jboss-modules 1.8.8.

dmlloyd:
We need to fix the CI issue before we can merge this PR.  The CI issue is likely due to a mistake in the test which assumes a certain test environment, and it would need one or two backports of fixes from later branches to correct.

spyrkob:
@dmlloyd i think the failures were caused by missing backport of MODULES-395. I opened #253 to fix that

PR #255, opened 2020-02-28
[MODULES-397] Removing deprecated -addindex , -modify and -logmodule command line options

ropalka:
https://issues.redhat.com/browse/MODULES-397


dmlloyd:
This should not be removed from 1.9 because as a policy we do not remove features in a micro (unless it's an exceptional case relating to security/CVE or a major bug).

ropalka:
Thanks for clarification.

PR #260, opened 2020-07-29
[MODULES-400] Automatically set the os.name if the -Djboss.modules.osname property has not already been specified when running on RHEL 6, RHEL 7, or RHEL 8

fjuma:
https://issues.redhat.com/browse/MODULES-400

### DIFF
--- a/src/main/java/org/jboss/modules/NativeLibraryResourceLoader.java
+++ b/src/main/java/org/jboss/modules/NativeLibraryResourceLoader.java
@@ -18,12 +18,19 @@
 
 package org.jboss.modules;
 
+import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * A base class for resource loaders which can load native libraries.
@@ -40,6 +47,16 @@ public class NativeLibraryResourceLoader extends AbstractResourceLoader {
         static final String CPU_ID;
         static final String ARCH_NAME;
         static final String[] NATIVE_SEARCH_PATHS;
+        static final Pattern OS_RELEASE_VERSION_ID_PATTERN = Pattern.compile("([0-9]+).*");
+        static final Pattern DISTRIBUTION_RELEASE_VERSION_PATTERN = Pattern.compile(".*\\s([0-9]+).*");
+        static final Pattern MAC_VERSION_PATTERN = Pattern.compile("([0-9]+\\.[0-9]+)\\.[0-9]+");
+        static final String OS_RELEASE_FILE = "/etc/os-release";
+        static final String FEDORA_RELEASE_FILE = "/etc/fedora-release";
+        static final String REDHAT_RELEASE_FILE = "/etc/redhat-release";
+        static final String ID = "ID=";
+        static final String VERSION_ID = "VERSION_ID=";
+        static final String RHEL = "rhel";
+        static final String FEDORA = "fedora";
 
         static {
             final Object[] strings = AccessController.doPrivileged(new PrivilegedAction<Object[]>() {
@@ -47,6 +64,7 @@ public class NativeLibraryResourceLoader extends AbstractResourceLoader {
                     // First, identify the operating system.
                     boolean knownOs = true;
                     String osName;
+                    String osVersion = null;
                     // let the user override it.
                     osName = System.getProperty("jboss.modules.os-name");
                     if (osName == null) {
@@ -58,8 +76,14 @@ public class NativeLibraryResourceLoader extends AbstractResourceLoader {
                             sysOs = sysOs.toUpperCase(Locale.US);
                             if (sysOs.startsWith("LINUX")) {
                                 osName = "linux";
+                                osVersion = getLinuxOSVersion();
                             } else if (sysOs.startsWith("MAC OS")) {
                                 osName = "macosx";
+                                String sysVersion = System.getProperty("os.version");
+                                Matcher m = MAC_VERSION_PATTERN.matcher(sysVersion);
+                                if (m.matches()) {
+                                    osVersion = m.group(1);
+                                }
                             } else if (sysOs.startsWith("WINDOWS")) {
                                 osName = "win";
                             } else if (sysOs.startsWith("OS/2")) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/modules/NativeLibraryResourceLoader.java
```diff
@@ -57,7 +60,13 @@
                         } else {
                             sysOs = sysOs.toUpperCase(Locale.US);
                             if (sysOs.startsWith("LINUX")) {
-                                osName = "linux";
+                                String sysVersion = System.getProperty("os.version");
+                                Matcher m = RHEL_PATTERN.matcher(sysVersion);
+                                if (m.matches()) {
+                                    osName = m.group(1); // el6, el7, or el8
```
dmlloyd: This feels wrong.  The ABI is still `linux`, so what is really going on here?  Are we selecting a different version of a library based on the OS version?  If so, maybe it makes more sense to have a more specific category and a fallback to a more general category; e.g. first search `linux-fc30-x86_64` then fall back to `linux-x86_64` if no match.  WDYT?

fjuma: This change is motivated by an issue that we've run into with the RHEL natives for WildFly OpenSSL. Previously, we used to have one Linux native for all RHEL platforms. So when running on a RHEL platform, `linux-x86_64` was added to the search path here and the WildFly OpenSSL native was able to be loaded successfully. We were recently asked by productization to have a separate Linux native for each RHEL platform, where each native is located in a directory of the form `el{6,7,8}-{cpuName}` (we could change the name of this directory if needed though). Since `System.getProperty("os.name")` will always return "linux" for any Linux distribution, it seems like we'd need to make use of `System.getProperty("os.version")` to determine the specific OS version (this returns something like "3.10.0-1127.el7.x86_64" on RHEL, not sure if the format is similar on all Linux distributions though). Is there a generic approach that we could use to create the specific category with the right OS version?

dmlloyd: I think that we should start with `<osName>-<version>-<cpu>` (e.g. `linux-el5-x86_64`) and fall back to the existing `<osName>-<cpu>` if the corresponding native library is not found in the OS-version-specific directory.  In the end RHEL is still Linux, so IMO we don't want to break compatibility there.  This should solve the productization requirement while also not breaking compatibility for things that are not controlled by productization.

That said, your version regex could easily be generalized to cover `fc` (Fedora) in addition to `el`.  And, it might be worth adding support for Mac OS's `<major>.<minor>` (leave off the micro) as well.

dmlloyd: I did some research on this and what I found was a little disheartening but not unsolvable.  It appears that on Linux at least, the `os.version` comes from the kernel - meaning if there's a custom kernel build, all of this detection logic will fail.

But there is a solution.  The [LSB specifies](https://linux.die.net/man/1/lsb_release) that there should be release information in `/etc`, and [SystemD also includes a useful file](http://0pointer.de/blog/projects/os-release.html) as well.  So we can reliable detect the Linux version in use, regardless of the kernel version, as follows:

* If `/etc/os-release` exists, we can parse it and get the `ID` and `VERSION_ID` fields to reliably know the distribution name and ID.
* If that fails, we can look in `/etc/` to locate distro-specific files, specifically `/etc/fedora-release`, `/etc/redhat-release` (which is present on both RHEL and Fedora, so it would have to be parsed), etc.  See [this link](https://www.unix.com/302075445-post4.html) for a forum post that might be useful.

fjuma: @dmlloyd Thanks very much for this info! Just checked the contents of `/etc/os-release` on Fedora - `ID` is `fedora` and `VERSION_ID` is 31. To form the OS-version-specific category, do you prefer `linux-fedora31-<cpu>` or `linux-fc31-<cpu>`?

Also checked the contents of `/etc/os-release` for RHEL - `ID` is `rhel` and `VERSION_ID` is `8.2`. To form the OS-version-specific category, do you prefer `linux-rhel8-<cpu>` or `linux-el8-<cpu>`?

dmlloyd: Good question.  I guess, to keep it simple, we should go with `fedora`/`rhel` since that's verbatim what comes out of that file.  That way it's probably already compatible with other Linux distributions out of the box.

fjuma: Thanks, @dmlloyd. I've now updated this to make use of the contents of `/etc/os-release`, `/etc/fedora-release`, and `/etc/redhat-release` to detect the Linux version in use. I also did some quick sanity checks on Fedora and RHEL 6, 7, and 8 to make sure that `getLinuxOSVersion()` returns the correct value (e.g., `fedora31`, `rhel{6,7,8}`).

src/main/java/org/jboss/modules/NativeLibraryResourceLoader.java
```diff
@@ -267,6 +303,84 @@
...
             NATIVE_SEARCH_PATHS = (String[]) strings[3];
         }
+
+        private static String getLinuxOSVersionFromOSReleaseFile() {
+            InputStream releaseFile;
+            try {
+                releaseFile = new FileInputStream(OS_RELEASE_FILE);
+            } catch (FileNotFoundException e) {
+                return null;
+            }
```
dmlloyd:
```suggestion
            try (InputStream releaseFile = new FileInputStream(OS_RELEASE_FILE)) {
```

I'd also add separate try-with-resources for each level of closeable, but only as a stylistic improvement.

src/main/java/org/jboss/modules/NativeLibraryResourceLoader.java
```diff
@@ -267,6 +303,84 @@
...
+                    return id + abbreviatedVersionId;
+                }
+            } catch (Exception e) {
+                return null;
+            }
+            return null;
+        }
+
+        private static String getLinuxOSVersionFromDistributionFile(String distributionFile) {
+            InputStream releaseFile;
```
dmlloyd: Same change here.

dmlloyd:
Other than the one comment, it looks fine.

src/main/java/org/jboss/modules/NativeLibraryResourceLoader.java
fjuma: Fixed.

src/main/java/org/jboss/modules/NativeLibraryResourceLoader.java
fjuma: Fixed.

dmlloyd:
OK this all looks good.  What branches would we need to merge this into?  Just 1.10/1.x?  Or will this need backporting?

fjuma:
Great. We just need this in 1.10/1.x. I'll create a PR for 1.x now. Thanks!

PR #261, opened 2020-07-30
[MODULES-400] [1.x] Automatically add an OS-version-specific category to the native search path for Linux and Mac OS and fallback to the previous generic category if the OS-version-specific category is not found

fjuma:
https://issues.redhat.com/browse/MODULES-400


dmlloyd:
Merged manually.  Thanks!

PR #262, opened 2020-08-12
Add GitHub actions

luck3y:
Signed-off-by: Ken Wills <kwills@redhat.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,48 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: JBoss Modules CI
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+jobs:
+  build-test-matrix:
+    name: JDK${{ matrix.jdk }}-${{ matrix.impl}}-${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        jdk: [11]
+        impl: [hotspot, openj9]
+    steps:
+    - uses: actions/checkout@v2
+    # test runs use -Djava8.home=... as well, so we need both JDKs
+    - name: Set up JDK 8
+      uses: AdoptOpenJDK/install-jdk@v1
+      with:
+        version: 8
+        impl: ${{ matrix.impl }}
+    - name: Store JDK8 env (Linux)
+      if: contains(matrix.os, 'ubuntu')
+      run: |
+        echo "::set-env name=JAVA8_HOME::$(echo $JAVA_HOME)"
+    - name: Store JDK8 env (Windows)
+      if: contains(matrix.os, 'windows')
+      run: |
+        echo "::set-env name=JAVA8_HOME::$env:JAVA_HOME"
+    - name: Set up JDK ${{ matrix.jdk }}
+      uses: AdoptOpenJDK/install-jdk@v1
+      with:
+        version: ${{ matrix.jdk }}
+        impl: ${{ matrix.impl }}
+    - name: Run Build & Test
+    # -D needs to be escaped in powershell, use --define= instead
+      run: mvn -U -B -fae clean install --define=java8.home=${{ env.JAVA8_HOME }}
+    - uses: actions/upload-artifact@v2
+      if: failure()
+      with:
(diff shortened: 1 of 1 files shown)


luck3y:
@dmlloyd Sorry if these are pestering you, I just have an irresistible urge to move them out of TC.

This one was a bit of a puzzle based on the need for JDK8 + whatever else is being tested due to windows env vars and powershell.

Oh, and also, since master wasn't currently building, I opened this against 1.x, LMK if it should be master first.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+
+name: JBoss Modules CI
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+jobs:
+  build-test-matrix:
+    name: JDK${{ matrix.jdk }}-${{ matrix.impl}}-${{ matrix.os }}
```
dmlloyd, line 12: This project is using the jboss-parent MR JAR mechanism so it should not be built using matrix.  There aren't any instructions for setting up GH actions for this on the JBoss Parent project, but it's basically the same process as it is for SmallRye Parent which is documented here: https://github.com/smallrye/smallrye-parent/#using-mr-jar-functions-with-github-actions.....

dmlloyd: Which I see you've essentially already done... let me read further.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+      if: contains(matrix.os, 'windows')
+      run: |
+        echo "::set-env name=JAVA8_HOME::$env:JAVA_HOME"
+    - name: Set up JDK ${{ matrix.jdk }}
+      uses: AdoptOpenJDK/install-jdk@v1
+      with:
+        version: ${{ matrix.jdk }}
+        impl: ${{ matrix.impl }}
+    - name: Run Build & Test
+    # -D needs to be escaped in powershell, use --define= instead
```
dmlloyd, line 42: :man_facepalming:

.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+    # test runs use -Djava8.home=... as well, so we need both JDKs
+    - name: Set up JDK 8
+      uses: AdoptOpenJDK/install-jdk@v1
+      with:
+        version: 8
+        impl: ${{ matrix.impl }}
+    - name: Store JDK8 env (Linux)
+      if: contains(matrix.os, 'ubuntu')
+      run: |
+        echo "::set-env name=JAVA8_HOME::$(echo $JAVA_HOME)"
```
dmlloyd, line 31: Why bother with setting an env var here?  The AdoptOpenJDK install job should set `JAVA_HOME_8` for you already.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+      run: |
+        echo "::set-env name=JAVA8_HOME::$env:JAVA_HOME"
+    - name: Set up JDK ${{ matrix.jdk }}
+      uses: AdoptOpenJDK/install-jdk@v1
+      with:
+        version: ${{ matrix.jdk }}
+        impl: ${{ matrix.impl }}
+    - name: Run Build & Test
+    # -D needs to be escaped in powershell, use --define= instead
+      run: mvn -U -B -fae clean install --define=java8.home=${{ env.JAVA8_HOME }}
```
dmlloyd, line 43:
```suggestion
      run: mvn -U -B -fae clean install --define=java8.home=${{ env.JAVA_HOME_8 }}
```

Assuming it works on Windows that is.  If so, you could delete the "Store JDK8 env" bits...

dmlloyd: It would be good to also add java9 and java10 TBH.

.github/workflows/ci.yml
luck3y: Well, yeah it does, then installing openjdk 11 overwrites it ;)

.github/workflows/ci.yml
luck3y: Oh, I see the TARGETs thing, I'll give that a go, thanks for the pointer.

dmlloyd:
Yeah `master` is EOL; the default branch is `1.x`.

.github/workflows/ci.yml
luck3y: +1 on the var rename, and yeah will add 9 & 10 as well

dmlloyd:
I mean you don't have to *close* it :)

luck3y:
Oh, I just closed it to open a test PR in my fork and make sure it worked ;)

luck3y:
I guess I'll have to reopen as a different PR, odd, thats usually fine even if I force push..

PR #267, opened 2020-11-05
[MODULES-402] Introducing Module.findServices() utility method

ropalka:
https://issues.redhat.com/browse/MODULES-402

### DIFF
--- a/src/main/java/org/jboss/modules/JDKSpecific.java
+++ b/src/main/java/org/jboss/modules/JDKSpecific.java
@@ -30,8 +30,12 @@ import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Collections;
 import java.util.Enumeration;
+import java.util.Iterator;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.ServiceLoader;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.jar.JarFile;
 
 import sun.reflect.Reflection;
@@ -183,6 +187,37 @@ final class JDKSpecific {
     static void addInternalPackages(final List<String> list) {
         list.add("sun.reflect.");
         list.add("jdk.internal.reflect.");
+    }
+
+    static <T> Iterable<T> findServices(final Class<T> serviceType, final Predicate<Class<?>> filter, final ClassLoader classLoader) {
+        final Iterator<T> delegate = ServiceLoader.load(serviceType, classLoader).iterator();
+        return new Iterable<T>() {
+            public Iterator<T> iterator() {
+                return new Iterator<T>() {
+                    T next = null;
+
+                    public boolean hasNext() {
+                        T next;
+                        while (this.next == null) {
+                            if (!delegate.hasNext()) return false;
+                            next = delegate.next();
+                            if (filter.test(next.getClass())) {
+                                this.next = next;
+                                return true;
+                            }
+                        }
+                        return true;
+                    }
+
+                    public T next() {
+                        if (!hasNext()) throw new NoSuchElementException();
+                        T next = this.next;
+                        this.next = null;
+                        return next;
+                    }
+                };
+            }
+        };
     }
 
     // === nested util stuff, non-API ===
--- a/src/main/java9/org/jboss/modules/JDKSpecific.java
+++ b/src/main/java9/org/jboss/modules/JDKSpecific.java
@@ -46,9 +46,13 @@ import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.ServiceLoader;
 import java.util.Set;
-import java.util.stream.Stream;
+import java.util.stream.Collectors;
+import java.util.function.Predicate;
 import java.util.jar.JarFile;
+import java.util.stream.Stream;
 
 /**
  * JDK-specific classes which are replaced for different JDK major versions.  This one is for Java 9 only.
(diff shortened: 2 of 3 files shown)


src/main/java/org/jboss/modules/JDKSpecific.java
```diff
@@ -185,6 +189,19 @@ static void addInternalPackages(final List<String> list) {
...
+        T provider;
+        while (i.hasNext()) {
+            provider = i.next();
+            if (provider != null && filter.test(provider.getClass())) {
+                providers.add(provider);
+            }
+        }
+        return providers;
+    }
+
```
dmlloyd: Other than `LinkedList` being a class that should *never* be used, I'd change to something like this:

```suggestion
        ServiceLoader<T> sl = ServiceLoader.load(serviceType, classLoader);
        return new Iterable<T>() {
            public Iterator<T> iterator() {
                final Iterator<T> delegate = sl.iterator();
                return new Iterator<T>() {
                    T next = null;
                    public boolean hasNext() {
                        while (next == null) {
                            if (! delegate.hasNext()) {
                                return false;
                            }
                            next = delegate.next();
                            if (filter.test(next.getClass())) {
                                this.next = next;
                                return true;
                            }
                        }
                        return true;
                    }
                    
                    public T next() {
                        if (! hasNext()) throw new NoSuchElementException();
                        T next = this.next;
                        this.next = null;
                        return next;
                    }
                }
            }
        }
    }

```

src/main/java9/org/jboss/modules/JDKSpecific.java
```diff
@@ -188,6 +191,10 @@ static void addInternalPackages(final List<String> list) {
         // none in Java 9+
     }
 
+    static <T> Iterable<T> findServices(final Class<T> serviceType, final Predicate<Class<?>> filter, final ClassLoader classLoader) {
+       return ServiceLoader.load(serviceType, classLoader).stream().filter(p -> filter.test(p.type())).map(ServiceLoader.Provider::get).collect(Collectors.toSet());
```
dmlloyd: Again don't return a collection here, instead use `.stream().iterator()` and then filter it like above.  I'm not sure how ServiceConfigurationErrors are handled in that case but it should be tested.

ropalka: We'll let the ServiceConfigurationErrors to be handled by users using this method.

src/main/java/org/jboss/modules/JDKSpecific.java
ropalka: Thanks for proposal, that's really better (because of lazyness)

ropalka:
I updated the PR @dmlloyd as you suggested, please review.

src/main/java/org/jboss/modules/JDKSpecific.java
```diff
@@ -185,6 +189,36 @@ static void addInternalPackages(final List<String> list) {
...
+                            if (filter.test(next.getClass())) {
+                                this.next = next;
+                                return true;
+                            }
+                        }
+                        return false;
+                    }
+
+                    public T next() {
+                        if (this.next == null) throw new NoSuchElementException();
```
dmlloyd: This should call `hasNext()`; it's allowed to call `next()` without calling `hasNext()`.

src/main/java/org/jboss/modules/JDKSpecific.java
```diff
@@ -185,6 +189,36 @@ static void addInternalPackages(final List<String> list) {
...
 
+    static <T> Iterable<T> findServices(final Class<T> serviceType, final Predicate<Class<?>> filter, final ClassLoader classLoader) {
+        final Iterator<T> delegate = ServiceLoader.load(serviceType, classLoader).iterator();
+        return new Iterable<T>() {
+            public Iterator<T> iterator() {
+                return new Iterator<T>() {
+                    T next = null;
+
+                    public boolean hasNext() {
+                        T next;
```
dmlloyd, line 200: This method should always return `true` if `this.next` is non-null.  Otherwise `hasNext()` will skip upstream elements.

src/main/java9/org/jboss/modules/JDKSpecific.java
```diff
@@ -188,6 +192,36 @@ static void addInternalPackages(final List<String> list) {
...
 
+    static <T> Iterable<T> findServices(final Class<T> serviceType, final Predicate<Class<?>> filter, final ClassLoader classLoader) {
+       final Iterator<ServiceLoader.Provider<T>> delegate = ServiceLoader.load(serviceType, classLoader).stream().iterator();
+       return new Iterable<T>() {
+            public Iterator<T> iterator() {
+                return new Iterator<T>() {
+                    T next = null;
+
+                    public boolean hasNext() {
+                        ServiceLoader.Provider<T> next;
```
dmlloyd, line 203: Same problem as above - if `this.next` is not `null` then `true` should be returned.

src/main/java9/org/jboss/modules/JDKSpecific.java
```diff
@@ -188,6 +192,36 @@ static void addInternalPackages(final List<String> list) {
...
+                            if (filter.test(next.type())) {
+                                this.next = next.get();
+                                return true;
+                            }
+                        }
+                        return false;
+                    }
+
+                    public T next() {
+                        if (this.next == null) throw new NoSuchElementException();
```
dmlloyd: Same problem as above - delegate to `hasNext()` always.

src/main/java/org/jboss/modules/JDKSpecific.java
ropalka: Right!

src/main/java/org/jboss/modules/JDKSpecific.java
ropalka: Thanks!

src/main/java9/org/jboss/modules/JDKSpecific.java
ropalka: Fixed

src/main/java9/org/jboss/modules/JDKSpecific.java
ropalka: Fixed

ropalka:
Thanks for valuable feedback @dmlloyd - it is fixed.

dmlloyd (approved):

PR #271, opened 2021-09-09
[MODULES-406] IllegalArgumentException when iterating modules if module-absent is present

rmartinc:
Issue: https://issues.redhat.com/browse/MODULES-406

Three little issues fixed when iterating modules if an `absent-module` is present:

1.  In the `LocalModuleFinder` class the returned `moduleSpec` can be null if it's an absent-module. Just adding the the null check.
2. When reading an absent-module the `required` property was not initialized OK if the xml version is 1.6+.
3. The `moduleName` passed to the `parseModuleAbsentContents` can be null and it should not be compared with the found one in that case (it seems that this check was added to the other parse methods but not for this one).

@ropalka Please review this PR when you have time.


dmlloyd (approved):

ropalka (approved):

ropalka:
Thanks @rmartinc !

PR #276, opened 2021-11-10
[MODULES-411] Nullpointers should not be dereferenced

boris-unckel:
Fixes https://issues.redhat.com/browse/MODULES-411


boris-unckel:
@ropalka Tests passed. Could you review and approve, please?

dmlloyd (approved):

PR #277, opened 2021-11-10
[MODULES-412] Fix conditional structure

boris-unckel:
Fixes https://issues.redhat.com/browse/MODULES-412

All branches in a conditional structure should not have exactly the same implementation


boris-unckel:
@ropalka This fix is easy and strange at the same time. Technically it's the same as before, just removing the if/else. Some methods have a distinction between `LIST_A_NAME` and something different, some not.

dmlloyd (approved):

dmlloyd:
Approved, however this change doesn't actually have any functional difference.

PR #282, opened 2021-11-11
[MODULES-417] "StandardCharsets" constants should be preferred

boris-unckel:
Fixes https://issues.redhat.com/browse/MODULES-417


ropalka (approved):

ropalka:
Thanks @boris-unckel !

PR #292, opened 2022-04-22
Introduce `module-info`

dmlloyd:
This one is for `2.x`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,9 @@
     <properties>
         <skip.compile>false</skip.compile>
         <version.org.wildfly.checkstyle-config>1.0.5.Final</version.org.wildfly.checkstyle-config>
-        <version.bridger>1.4.Final</version.bridger>
+        <version.bridger>1.6.Final</version.bridger>
+        <version.module-info.plugin>1.2</version.module-info.plugin>
+        <version.surefire.plugin>3.0.0-M6</version.surefire.plugin>
 
         <jdk.min.version>9</jdk.min.version>
     </properties>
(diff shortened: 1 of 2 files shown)


jamezp (approved):

ropalka:
What's the main idea / usecase behind this PR @dmlloyd ?

After I built JBoss Modules with this PR included I can see there are
6 'uses' directives defined in generated moule-info.class:

# explicitly specified via module-info.yml
  * uses java.util.logging.LogManager;
  * uses java.util.prefs.PreferencesFactory;
  * uses javax.management.MBeanServerBuilder;
# added via bytecode inspection
  * uses java.lang.SecurityManager;
  * uses java.security.Provider;
  * uses org.jboss.modules.PreMain;

According to [1] 'uses' directives specify services for which
the module may discover providers via j.u.ServiceLoader.

JBoss Modules API provides these three methods:
 * Module.findServices()
 * Module.loadService()
 * Module.loadServiceDirectly()

These three methods are kind of a 'dynamic' capability to lookup
various services in provided / defined classloaders. However since
ServiceLoader.load() methods are @CallerSensitive these methods will not
work if jboss-modules.jar is put on the module path.
j.u.ServiceLoader calls will fail for services not specified
in generated module-info.class.

[1] https://docs.oracle.com/javase/specs/jls/se9/html/jls-7.html#jls-7.7.3

ropalka:
Partial answer was posted here: https://github.com/jboss-logging/jboss-logmanager/pull/362#issuecomment-1108665903

ropalka (approved):
Approving since this PR is an incremental change of a more complex effort.

PR #294, opened 2022-09-23
[MODULES-426] Switch to JDK11 by default

ropalka:
https://issues.redhat.com/browse/MODULES-426


dmlloyd (commented):
What is the purpose of adding these packages? The only time that these files should be used is when running under Java 8, correct? So, it is expected that these paths would be missing (and other paths that are not found on Java 8). Beyond Java 8, all of the paths will be added automatically and correctly.

dmlloyd (approved):

ropalka:
Thanks for review @dmlloyd !

PR #298, opened 2023-01-13
[MODULES-435] Improve performance of repeat ModuleIdentifier toString and caclulateHashCode calls

aogburn:
https://issues.redhat.com/browse/MODULES-435

### DIFF
--- a/src/main/java/org/jboss/modules/ModuleIdentifier.java
+++ b/src/main/java/org/jboss/modules/ModuleIdentifier.java
@@ -21,6 +21,8 @@ package org.jboss.modules;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.Serializable;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * A unique identifier for a module within a module loader.
@@ -40,8 +42,12 @@ public final class ModuleIdentifier implements Serializable {
 
     private static final String DEFAULT_SLOT = "main";
 
+    private static final Map<String, ModuleIdentifier> identifierMap = new ConcurrentHashMap<>();
+
     private final String name;
     private final String slot;
+
+    private transient String toString;
 
     private transient volatile int hashCode;
 
@@ -118,8 +124,12 @@ public final class ModuleIdentifier implements Serializable {
      */
     @Override
     public String toString() {
+        if (toString != null) {
+            return toString;
+        }
         final String escapeName = escapeName(name);
-        return slot.equals(DEFAULT_SLOT) ? escapeName : escapeName + ":" + escapeSlot(slot);
+        toString = slot.equals(DEFAULT_SLOT) ? escapeName : escapeName + ":" + escapeSlot(slot);
+        return toString;
     }
 
     private void readObject(ObjectInputStream ois) throws ClassNotFoundException, IOException {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/modules/ModuleIdentifier.java
```diff
@@ -40,9 +42,13 @@ public final class ModuleIdentifier implements Serializable {
 
     private static final String DEFAULT_SLOT = "main";
 
+    private static Map<String, ModuleIdentifier> identifierMap = new ConcurrentHashMap<>();
```
dmlloyd: We definitely do _not_ want a static global cache of identifiers. (But if we did, we would want it to be `final`.)

src/main/java/org/jboss/modules/ModuleIdentifier.java
```diff
@@ -40,9 +42,13 @@ public final class ModuleIdentifier implements Serializable {
...
+
     private final String name;
     private final String slot;
 
+    private String escapedString;
+
```
dmlloyd: Idiomatically, we usually would call this field `toString` in order to match up with the method from which it is used. And, it would need to be `transient`.

dmlloyd (changes requested):

src/main/java/org/jboss/modules/ModuleIdentifier.java
aogburn: Any other ideas for improving potential degradation from repeat ModuleIdentifier.fromString calls aside from an identifier cache to return repeat values?

dmlloyd: We could intern the assembled string, which should eliminate the bulk of the heap retention.

But I think the best thing we could do would be to eliminate all usages of `ModuleIdentifier` from upstream code.

aogburn: After MODULES-256, the ModuleIdentifier deprecation changes add processing time overhead in potentially repeating fromString and toString calls so the concern isn't really with heap for String intern to help.  It's unfortunate that deprecation then introduced a varying performance degradation in anything using it and we may not address it in JBoss Modules itself.  I filed WFCORE-6211 to at least initially remove the problem ModuleIdentifier use in the ServiceModuleLoader that causes the specific start up slowness I've seen.  But it may be a lot of large sweeping changes required upstream to ensure ModuleIdentifier is not used anywhere to not see its fromString/toString overhead.

dmlloyd: I added a comment to that issue, but generally speaking I think removing usages is the best solution, followed by adding caching as needed to the individual call sites (which might be a useful solution in certain cases where compatibility is a problem).

aogburn: I've just seen JBoss modules itself as the call site from Module.addExportedPaths / ModuleLoader.preloadModule as described in MODULES-435.  So we'd move caching if done in this case from the ModuleIdentifier to the ModuleLoader?  Or override preloadmodule(String) in wildfly's ServiceModuleLoader to cache?  But then if ServiceModuleLoader (or some other implementation) can provide its own preloadModule(String) method, we're likely already addressing the issue as Module.addExportedPaths would call ServiceModuleLoader.preloadModule(String) instead of ModuleLoader.preloadModule(String) and its resulting from/toStrings.

dmlloyd: By changing the override to the supported (non-deprecated) method, the caller will no longer be `preloadModule`. There is special logic in the constructor of `ModuleLoader` which only calls the deprecated `ModuleIdentifer`-based `preloadModule` when it is overridden by the subclass, in order to emulate the legacy behavior where modules were loaded by identifier; when it isn't, that call path is never used, thus no `ModuleIdentifer` will ever be constructed, thus no need for the cache.

dmlloyd:
This change *should* no longer be needed based on upstream changes. If a problem persists, we can follow up on #726.

PR #301, opened 2023-03-02
Large deployments cause a corresponding number of allocations during canonicalize

chrisruffalo:
With EAP 7.2.9 we ran into an issue with having a lot of GC during startup. This PR adds a few strategies to reduce allocations during canonicalizing paths in response to issues with large deployments.

Before:
image

After:
image

As you can see, prior to the change, there were something like 250k allocations. After the change there were seeing closer to 50k. On another instance we were seeing around 65k allocations before and now see around 34k.

While these changes didn't have the affect we were hoping for on startup times we do see a lot less heap pressure and GC events during startup.

I am not sure if this implementation is 100% correct. I'm also not sure that reusing the buffers is required. The most impact seemed to come from just inserting more cases where we know we don't need to do further canonicalization. What I do know is that this is an issue for some small subset of deployments.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -284,6 +284,16 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <version>1.36</version>
+        </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>1.36</version>
+        </dependency>
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>4.11</version>
--- a/pom.xml
+++ b/pom.xml
@@ -35,7 +35,7 @@
     <licenses>
         <license>
             <name>Apache License 2.0</name>
-            <url>http://:y.jboss.org/licenses/apache-2.0.txt</url>
+            <url>http://repository.jboss.org/licenses/apache-2.0.txt</url>
             <distribution>repo</distribution>
         </license>
         <license>
--- a/pom.xml
+++ b/pom.xml
@@ -93,6 +93,7 @@
                             <execution>
                                 <id>default-test</id>
                                 <configuration>
+                                    <argLine>-Xmx2048m -Xms2048m</argLine>
                                     <includes>
                                         <include>**/BenchmarkSuite*</include>
                                     </includes>
--- a/pom.xml
+++ b/pom.xml
@@ -35,7 +35,7 @@
     <licenses>
         <license>
             <name>Apache License 2.0</name>
-            <url>http://repository.jboss.org/licenses/apache-2.0.txt</url>
+            <url>http://:y.jboss.org/licenses/apache-2.0.txt</url>
             <distribution>repo</distribution>
         </license>
         <license>
@@ -74,6 +74,31 @@
                                 </goals>
                                 <configuration>
                                     <jvm>${java8.home}/bin/java</jvm>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+        <profile>
+            <id>benchmark</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>default-test</id>
+                                <configuration>
+                                    <includes>
+                                        <include>**/BenchmarkSuite*</include>
+                                    </includes>
+                                    <excludes>
+                                        <exclude>**/*Test*</exclude>
+                                    </excludes>
                                 </configuration>
                             </execution>
                         </executions>
@@ -140,7 +165,7 @@
             <!-- Compiler -->
             <plugin>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.7.0-jboss-1</version>
+                <version>3.8.1-jboss-1</version>
                 <executions>
                     <execution>
                         <id>default-compile</id>
@@ -192,9 +217,6 @@
                      <!-- Modules ignores Class-Path -->
                     <useManifestOnlyJar>false</useManifestOnlyJar>
                     <printSummary>true</printSummary>
-                    <includes>
-                        <include>**/*Test.java</include>
-                    </includes>
                     <forkMode>always</forkMode>
                 </configuration>
                 <executions>
@@ -202,6 +224,12 @@
                         <id>default-test</id>
                         <configuration>
                             <classesDirectory>${project.build.directory}/classes/META-INF/versions/9</classesDirectory>
+                            <excludes>
+                                <exclude>**/*Benchmark*</exclude>
+                            </excludes>
+                            <includes>
+                                <include>**/*Test.java</include>
+                            </includes>
                             <additionalClasspathElements>
                                 <additionalClasspathElement>${project.build.directory}/classes</additionalClasspathElement>
                             </additionalClasspathElements>
--- a/src/main/java/org/jboss/modules/PathUtils.java
+++ b/src/main/java/org/jboss/modules/PathUtils.java
@@ -227,14 +227,16 @@ public final class PathUtils {
                         // not reached!
                     }
 
-                    // look for the next boundary
+                    // look for the next potential boundary point
                     int newE = e > 0 ? path.lastIndexOf('/', e - 1) : -1;
 
-                    // this is a slight optimization that reduces the number of copies for strings that are
-                    // checking paths that need fewer modifications but it can only be done when no skipping
-                    // needs to be performed and it leads to an earlier "out" / allows longer strings to get
-                    // an early out on the zero-allocation path
-                    if (skip < 1) {
+                    if (skip > 0) {
+                        skip--;
+                    } else {
+                        // this is a slight optimization that reduces the number of copies for strings that are
+                        // checking paths that need fewer modifications but it can only be done when no skipping
+                        // needs to be performed and it leads to an earlier "out" / allows longer strings to get
+                        // an early out on the zero-allocation path
                         do {
                             if (newE < 1) {
                                 break;
@@ -246,28 +248,30 @@ public final class PathUtils {
                                 break;
                             }
                         } while (newE != -1);
-                    }
 
-                    // create segment length for the region that will be copied, which should be from the segment end
-                    // to wherever we found a boundary that cannot be crossed
-                    final int segmentLength = e - newE - 1;
+                        // create segment length for the region that will be copied, which should be from the segment end
+                        // to wherever we found a boundary that cannot be crossed
+                        final int segmentLength = e - newE - 1;
 
-                    // at this point we are attempting to copy the entire string so just return which avoids any sort
-                    // of allocation for non-modified strings
-                    if (segmentLength == length) {
-                        return path;
-                    }
-
-                    if (skip > 0) {
-                        skip--;
-                    } else {
-                        if (targetBuf == null) {
-                            targetBuf = getBuffer(length);
+                        // at this point we are attempting to copy the entire string so just return which avoids any sort
+                        // of allocation for non-modified strings
+                        if (segmentLength == length) {
+                            return path;
                         }
-                        if (state == 3) {
-                            targetBuf[a--] = '/';
+
+                        if (segmentLength != 0) {  // not sure if this realistically skips anything
+                            if (targetBuf == null) {
+                                targetBuf = getBuffer(length);
+                            }
+                            if (state == 3) {
+                                targetBuf[a--] = '/';
+                            }
+//                            if (pathBuf == null) {
+//                                pathBuf = path.toCharArray();
+//                            }
+//                            System.arraycopy(pathBuf, newE + 1, targetBuf, (a -= segmentLength) + 1, segmentLength);
+                            path.getChars(newE + 1, e, targetBuf, (a -= segmentLength) + 1);
                         }
-                        path.getChars(newE + 1, e, targetBuf, (a -= segmentLength) + 1);
                     }
                     state = 0;
                     i = newE + 1;
@@ -276,19 +280,29 @@ public final class PathUtils {
                 }
             }
         }
-
+        // if we exit on a state of 3 then the path is absolute
         if (state == 3) {
+            // if the string is _only_ the absolute character
+            // do not allocate a new string
             if (length - a - 1 == 1) {
                 return "/";
             }
+            // we should not get here because if we do it somehow means we
+            // have an empty string but nothing in it that we are going to
+            // prepend to (remove?)
             if (targetBuf == null) {
                 targetBuf = getBuffer(length);
             }
+            // prepend the absolute start and also move the starting position of
+            // the buffer to the left by one.
             targetBuf[a--] = '/';
         }
+        // if the string has been cut down to a length of 0 then
+        // do not allocate anything and return an empty string
         if (length - a - 1 == 0) {
             return "";
         }
+        // no change was made so return the original path
         if (targetBuf == null) {
             return path;
         }
--- a/src/main/java/org/jboss/modules/PathUtils.java
+++ b/src/main/java/org/jboss/modules/PathUtils.java
@@ -268,6 +268,11 @@ public final class PathUtils {
     }
 
 
+    private static boolean isSepChar(char ch) {
+        // on non-Windows, the JIT *should* eliminate the second check as redundant
+        return ch == '/' || ch == File.separatorChar;
+    }
+
     /**
      * Canonicalize the given path.  Removes all {@code .} and {@code ..} segments from the path.
      *
@@ -275,14 +280,29 @@ public final class PathUtils {
      * @return the canonical path
      */
     public static String canonicalize(String path) {
-        if (path.trim().isEmpty()) {
+        // is null check here even needed?
+        if (path == null || path.isEmpty()) {
             return path;
         }
 
-        // represents a state transitions that will be found via the state machine
-        // but if these patterns aren't contained in the string they can't be found
-        // via the state machine. (we know it already contains at least "." at this point.)
-        if (!path.contains("/.") && !path.contains("./")) {
+        boolean found = false;
+        int idx = path.indexOf('.'); // maybe reuse the result of the above comparison for efficiency
+        // shortcut if string is "." or ".."
+        if (idx == 0 && path.length() == 1 || path.length() == 2 && path.charAt(idx+1) == '.') {
+            return "";
+        }
+        if (idx == -1) {
+            return path;
+        }
+        do {
+            // represents a state transitions that will be found via the state machine
+            // but if these patterns aren't contained in the string they can't be found
+            // via the state machine. (we know it already contains at least "." at this point.)
+            if (path.length() > idx + 1 && isSepChar(path.charAt(idx + 1)) || idx > 0 && isSepChar(path.charAt(idx - 1))) {
+                found = true;
+            }
+        } while (idx + 1 < path.length() && (idx = path.indexOf('.', idx + 1)) != -1);
+        if (!found) {
             return path;
         }
 
--- a/src/main/java/org/jboss/modules/PathUtils.java
+++ b/src/main/java/org/jboss/modules/PathUtils.java
@@ -35,7 +35,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 
 import org.jboss.modules.filter.PathFilter;
 
@@ -47,7 +46,7 @@ import org.jboss.modules.filter.PathFilter;
 public final class PathUtils {
 
     static final int MIN_LENGTH = 256;
-    private static final ThreadLocal<char[]> CHAR_BUFFER_CACHE = new ThreadLocal<>();
+    static final ThreadLocal<char[]> CHAR_BUFFER_CACHE = new ThreadLocal<>();
 
     private PathUtils() {
     }
@@ -163,14 +162,17 @@ public final class PathUtils {
         char[] current = CHAR_BUFFER_CACHE.get();
         if(current == null || current.length < length) {
             // since we can't shift left from the max value it will hold at the max value
-            // it uses the outer Math.max to belt-and-suspenders out of a lot of corner cases that end up negative.
+            // the length is checked against the min value and the min value is just returned if it is too low
             // the shifting logic finds out how many powers of two we need to shift left to get to the next power of two
             // requesting a size that is already a power of two will give you that same power of two
             // because we get the same power of two we use a Math.max on the inner value to make sure we always meet the minimum
-            int nexLength = length == Integer.MAX_VALUE ? Integer.MAX_VALUE : Math.max(MIN_LENGTH, 1 << (Integer.SIZE - Integer.numberOfLeadingZeros(Math.max(length, MIN_LENGTH) - 1)));
-            CHAR_BUFFER_CACHE.set(new char[nexLength]);
+            final int nextLength = length == Integer.MAX_VALUE ? Integer.MAX_VALUE
+                                : length <= MIN_LENGTH ? MIN_LENGTH
+                                : 1 << (Integer.SIZE - Integer.numberOfLeadingZeros(length - 1));
+            current = new char[nextLength];
+            CHAR_BUFFER_CACHE.set(current);
         }
-        return CHAR_BUFFER_CACHE.get();
+        return current;
     }
 
     /**
@@ -187,6 +189,9 @@ public final class PathUtils {
         // 2 - got two .
         // 3 - got /
         int state = 0;
+        if (length == 0) {
+            return path;
+        }
         char[] targetBuf = null;
         // string segment end exclusive
         int e = length;
@@ -211,10 +216,10 @@ public final class PathUtils {
                 }
                 case '.': {
                     inner: switch (state) {
-                        case 0: state = 3; e = i; break outer;
-                        case 1: state = 3; e = i; break outer;
-                        case 2: state = 3; e = i; skip ++; break outer;
-                        case 3: e = i; break outer;
+                        case 0: state = 1; break outer;
+                        case 1: state = 2; break outer;
+                        case 2: break inner; // emit!
+                        case 3: state = 1; break outer;
                         default: throw new IllegalStateException();
                     }
                     // fall thru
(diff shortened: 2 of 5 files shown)


src/main/java/org/jboss/modules/PathUtils.java
```diff
@@ -144,23 +148,52 @@ public static String fileNameOfPath(String path) {
...
      * @param path the relative or absolute possibly non-canonical path
      * @return the canonical path
      */
     public static String canonicalize(String path) {
+        if (path.trim().isEmpty() || (!path.contains("..") && !path.contains("."))) {
+            return path;
+        }
+        if (path.contains(".") && !path.contains("/.") && !path.contains("./")) {
+            return path;
+        }
```
dmlloyd: I would expect that these five whole-string scans would be more costly than just running the single-pass scan of the canonicalization routine, assuming that the changes to lazily acquire (or even just lazily allocate) the target buffer are retained. I would leave these two `if` conditions off.

src/main/java/org/jboss/modules/PathUtils.java
```diff
@@ -174,8 +207,9 @@ public static String canonicalize(String path) {
             outer: switch (c) {
                 case '/': {
                     inner: switch (state) {
-                        case 0: state = 3; e = i; break outer;
-                        case 1: state = 3; e = i; break outer;
+                        case 0:
+                        case 1:
+                            state = 3; e = i; break outer;
```
dmlloyd: I'd prefer to keep this as it was for readability, or at least keep the `case 1` line as it is and change `case 0` to be `case 0: // fall through`. I'd like a similar change to the other two sites.

dmlloyd (commented):
Good analysis and overall I like the change, but I have a couple of suggestions.

src/main/java/org/jboss/modules/PathUtils.java
chrisruffalo: I'll revert this. It's an artifact of me just clicking around and listening to the IDE.

src/main/java/org/jboss/modules/PathUtils.java
chrisruffalo: I had a JMH benchmark that I didn't want to introduce to the code base but my read of the test results was that it was quite a bit faster. I think we can also cut down the number of `contains` calls by a few and still get the same affect. 

I'll bring the benchmark back into this MR so we can compare. There's a good chance I'm reading it wrong.

I also need to run it again with just the changes for the buffer allocation. I couldn't get the JFR recordings off the remote host and so I'll do it again at some other point. I added the contains checks in a fit of pique because I wasn't seeing the reduction in deployment time I wanted.

chrisruffalo: I've reorganized the results a bit but here's the summary:
```
Benchmark                                      Mode  Cnt  Score    Error  Units
PathUtilsBenchmarkTest.changeString            avgt    4  0.136 �  0.004  us/op
PathUtilsBenchmarkTest.directChangeString      avgt    4  0.136 �  0.001  us/op

PathUtilsBenchmarkTest.manyChangeString        avgt    4  0.088 �  0.002  us/op
PathUtilsBenchmarkTest.directManyChangeString  avgt    4  0.088 �  0.001  us/op

PathUtilsBenchmarkTest.noChangeString          avgt    4  0.003 �  0.001  us/op
PathUtilsBenchmarkTest.directNoChangeString    avgt    4  0.130 �  0.003  us/op

PathUtilsBenchmarkTest.withDotButFine          avgt    4  0.003 �  0.001  us/op
PathUtilsBenchmarkTest.directWithDotButFine    avgt    4  0.054 �  0.005  us/op
```

In short there isn't much difference between ones that go through the if blocks and ones that don't but the one's that get out early are much much faster.

dmlloyd: With a difference like this, I guess that JIT might be coming into play, or perhaps the tight loops in `contains` are somehow optimized better than the state machine loop of the main routine.

dmlloyd: I wonder if replacing the `switch` with `if` would help here. The cases are few enough, and the JIT cannot profile `switch` like it can with `if`.

chrisruffalo: I think the method itself is fast enough it's just allocation pressure that slows it down. I'm also worried about the profiling and the fact that JMH does try and consume the value but the compiler might be too smart. I'm adding a few more test cases for the benchmark as well.

chrisruffalo: These changes really affect `isChild` as well, which is the most common entry point into `canonicalize` during the boot up process (from what i am reading). I was looking at that method to see if there were any gains to be made there but I can't see any way to get around modifying the string or anything like that. I'll probably create a benchmark case for that as well but that's probably going too far.

src/main/java/org/jboss/modules/PathUtils.java
```diff
@@ -274,6 +265,27 @@ public static String canonicalize(String path) {
...
     }
 
+    /**
+     * Canonicalize the given path.  Removes all {@code .} and {@code ..} segments from the path.
+     *
+     * @param path the relative or absolute possibly non-canonical path
+     * @return the canonical path
+     */
+    public static String canonicalize(String path) {
+        if (path.trim().isEmpty() || !path.contains(".")) {
```
dmlloyd: I'd try two things here:

* Drop the `trim()` part. If the path does not consist of only whitespace and does have leading or trailing whitespace, this will cause a new `String` allocation which is immediately discarded.
* Try changing `!path.contains(".")` to `path.indexOf('.') == -1`, which should eliminate a nested loop and some other extra work inside of `String`.

src/main/java/org/jboss/modules/PathUtils.java
```diff
@@ -274,6 +265,27 @@ public static String canonicalize(String path) {
...
+            return path;
+        }
+
+        // represents a state transitions that will be found via the state machine
+        // but if these patterns aren't contained in the string they can't be found
+        // via the state machine. (we know it already contains at least "." at this point.)
+        if (!path.contains("/.") && !path.contains("./")) {
```
dmlloyd: Here maybe try something like:

```java
int idx = path.indexOf('.'); // maybe reuse the result of the above comparison for efficiency
if (! (path.length() > idx + 1 && isSepChar(path.charAt(idx + 1)) || idx > 0 && isSepChar(path.charAt(idx - 1))) {
```

With:

```
private static boolean isSepChar(char ch) {
    // on non-Windows, the JIT *should* eliminate the second check as redundant
    return ch == '/' || ch == File.separatorChar;
}

Comment (dmlloyd) [reply src/main/java/org/jboss/modules/PathUtils.java]: Also you'd better account for the strings `"."` and `".."`, which do need to be canonicalized but do not contain `/`.

File: src/main/java/org/jboss/modules/PathUtils.java (hunk 1 above)
Comment (chrisruffalo) [reply src/main/java/org/jboss/modules/PathUtils.java]: Ok, so this is a lot but I've added the GC profiling to JMH benchmarking and extended the benchmark duration. It takes a minute to run and can be accessed with the `benchmark` profile. I have denoted the original algorithm, a direct call without filtering ("direct"), and the pre-checking methods in the table below.

| Case | Metric | Original | Direct | Pre-Checking |
|---|---|---|---|---|
| No Change String | us/op | 0.086 | 0.093 | 0.012 |
| | allocation (MB/sec) | 8536 | 3609 | 0.001 |
| | allocation (B/op) | 192 | 88 | 10^-6 |
| Changed String | us/op | 0.104 | 0.111 | 0.122 |
| | allocation (MB/sec) | 7072 | 2740 | 2496 |
| | allocation (B/op) | 192 | 80 | 80 |
| Many Change String | us/op | 0.070 | 0.063 | 0.066 |
| | allocation (MB/sec) | 8264 | 2906 | 2770 |
| | allocation (B/op) | 152 | 48 | 48 |
| With Dot But No Change | us/op | 0.063 | 0.040 | 0.009 |
| | allocation (MB/sec) | 9185 | 6785 | 0.001 |
| | allocation (B/op) | 152 | 72 | 10^-6 |
| Long String With Change | us/op | 1.433 | 1.635 | 1.571 |
| | allocation (MB/sec) | 3278 | 1007 | 1048 |
| | allocation (B/op) | 1232 | 432 | 432 |
| Long String With No Change | us/op | 1.536 | 1.610 | 1.684 |
| | allocation (MB/sec) | 3100 | 1042 | 996 |
| | allocation (B/op) | 1248 | 440 | 440 |

If you look at the long string with a change (390 characters that doesn't have a `/.` until the end) you can see in the results that the difference between the original and pre-checking is around 14% increase. I think the JVM is outsmarting us or something though because the _direct_ benchmarks are slower than the pre-checking benchmarks sometimes. 

The allocation rate is really what makes this worthwhile anyway. A lot of these test cases have 30% of the allocations compared to the original.

The real question is if a 10-14% decrease in speed is worth it for a large speedup (~7x) when canonicalization is definitely not needed. I don't know if I'm qualified to answer that. We also have to pay for the interaction with the ThreadLocal which is around 5% in speed over the original.

File: src/main/java/org/jboss/modules/PathUtils.java (hunk 4 above)
Comment (chrisruffalo) [reply src/main/java/org/jboss/modules/PathUtils.java]: Ah this makes sense, I will need to add some test cases to catch things like this. I will add some of these cases and then update the code.

Comment (chrisruffalo) [reply src/main/java/org/jboss/modules/PathUtils.java]: I found some bugs, I'm using the code you suggested now and came up with some better test cases. I will benchmark again (takes about 20-30min).

Comment (chrisruffalo) [reply src/main/java/org/jboss/modules/PathUtils.java]:
```
Benchmark                                                            Mode  Cnt     Score      Error   Units

# META-INF/application.properties
PathUtilsBenchmarkTest.withDotButFine                                avgt    5     0.011 ±    0.001   us/op
PathUtilsBenchmarkTest.withDotButFine:·gc.alloc.rate                 avgt    5     0.002 ±    0.012  MB/sec
PathUtilsBenchmarkTest.withDotButFine:·gc.alloc.rate.norm            avgt    5    ≈ 10⁻⁶               B/op
PathUtilsBenchmarkTest.withDotButFine:·gc.count                      avgt    5       ≈ 0             counts
PathUtilsBenchmarkTest.directWithDotButFine                          avgt    5     0.060 ±    0.001   us/op
PathUtilsBenchmarkTest.directWithDotButFine:·gc.alloc.rate           avgt    5  9168.130 ±  180.717  MB/sec
PathUtilsBenchmarkTest.directWithDotButFine:·gc.alloc.rate.norm      avgt    5    72.000 ±    0.001    B/op
PathUtilsBenchmarkTest.directWithDotButFine:·gc.count                avgt    5  1004.000             counts
PathUtilsBenchmarkTest.directWithDotButFine:·gc.time                 avgt    5   674.000                 ms
PathUtilsBenchmarkTest.originalWithDotButFine                        avgt    5     0.164 ±    0.009   us/op
PathUtilsBenchmarkTest.originalWithDotButFine:·gc.alloc.rate         avgt    5  7086.643 ±  386.548  MB/sec
PathUtilsBenchmarkTest.originalWithDotButFine:·gc.alloc.rate.norm    avgt    5   152.000 ±    0.001    B/op
PathUtilsBenchmarkTest.originalWithDotButFine:·gc.count              avgt    5   580.000             counts
PathUtilsBenchmarkTest.originalWithDotButFine:·gc.time               avgt    5  1167.000                 ms

# /this/./path/has/a/../need/to/be/canonicalized
PathUtilsBenchmarkTest.changeString                                  avgt    5     0.130 ±    0.001   us/op
PathUtilsBenchmarkTest.changeString:·gc.alloc.rate                   avgt    5  4680.515 ±   47.435  MB/sec
PathUtilsBenchmarkTest.changeString:·gc.alloc.rate.norm              avgt    5    80.000 ±    0.001    B/op
PathUtilsBenchmarkTest.changeString:·gc.count                        avgt    5   919.000             counts
PathUtilsBenchmarkTest.changeString:·gc.time                         avgt    5   496.000                 ms
PathUtilsBenchmarkTest.directChangeString                            avgt    5     0.112 ±    0.001   us/op
PathUtilsBenchmarkTest.directChangeString:·gc.alloc.rate             avgt    5  5433.140 ±   46.795  MB/sec
PathUtilsBenchmarkTest.directChangeString:·gc.alloc.rate.norm        avgt    5    80.000 ±    0.001    B/op
PathUtilsBenchmarkTest.directChangeString:·gc.count                  avgt    5   932.000             counts
PathUtilsBenchmarkTest.directChangeString:·gc.time                   avgt    5   514.000                 ms
PathUtilsBenchmarkTest.originalChangeString                          avgt    5     0.173 ±    0.024   us/op
PathUtilsBenchmarkTest.originalChangeString:·gc.alloc.rate           avgt    5  8485.424 ± 1181.296  MB/sec
PathUtilsBenchmarkTest.originalChangeString:·gc.alloc.rate.norm      avgt    5   192.000 ±    0.001    B/op
PathUtilsBenchmarkTest.originalChangeString:·gc.count                avgt    5   764.000             counts
PathUtilsBenchmarkTest.originalChangeString:·gc.time                 avgt    5   870.000                 ms

# /this/path/has/no/need/to/be/canonicalized
PathUtilsBenchmarkTest.noChangeString                                avgt    5     0.012 ±    0.001   us/op
PathUtilsBenchmarkTest.noChangeString:·gc.alloc.rate                 avgt    5     0.002 ±    0.012  MB/sec
PathUtilsBenchmarkTest.noChangeString:·gc.alloc.rate.norm            avgt    5    ≈ 10⁻⁶               B/op
PathUtilsBenchmarkTest.noChangeString:·gc.count                      avgt    5       ≈ 0             counts
PathUtilsBenchmarkTest.directNoChangeString                          avgt    5     0.094 ±    0.003   us/op
PathUtilsBenchmarkTest.directNoChangeString:·gc.alloc.rate           avgt    5  7141.097 ±  249.419  MB/sec
PathUtilsBenchmarkTest.directNoChangeString:·gc.alloc.rate.norm      avgt    5    88.000 ±    0.001    B/op
PathUtilsBenchmarkTest.directNoChangeString:·gc.count                avgt    5  1009.000             counts
PathUtilsBenchmarkTest.directNoChangeString:·gc.time                 avgt    5   617.000                 ms
PathUtilsBenchmarkTest.originalNoChangeString                        avgt    5     0.189 ±    0.051   us/op
PathUtilsBenchmarkTest.originalNoChangeString:·gc.alloc.rate         avgt    5  7794.562 ± 2195.670  MB/sec
PathUtilsBenchmarkTest.originalNoChangeString:·gc.alloc.rate.norm    avgt    5   192.000 ±    0.001    B/op
PathUtilsBenchmarkTest.originalNoChangeString:·gc.count              avgt    5   713.000             counts
PathUtilsBenchmarkTest.originalNoChangeString:·gc.time               avgt    5  1104.000                 ms

# /../../../../../.././.thing/../../././../
PathUtilsBenchmarkTest.manyChangeString                              avgt    5     0.124 ±    0.003   us/op
PathUtilsBenchmarkTest.manyChangeString:·gc.alloc.rate               avgt    5  2959.892 ±   82.816  MB/sec
PathUtilsBenchmarkTest.manyChangeString:·gc.alloc.rate.norm          avgt    5    48.000 ±    0.001    B/op
PathUtilsBenchmarkTest.manyChangeString:·gc.count                    avgt    5  1058.000             counts
PathUtilsBenchmarkTest.manyChangeString:·gc.time                     avgt    5   337.000                 ms
PathUtilsBenchmarkTest.directManyChangeString                        avgt    5     0.070 ±    0.001   us/op
PathUtilsBenchmarkTest.directManyChangeString:·gc.alloc.rate         avgt    5  5207.492 ±  103.130  MB/sec
PathUtilsBenchmarkTest.directManyChangeString:·gc.alloc.rate.norm    avgt    5    48.000 ±    0.001    B/op
PathUtilsBenchmarkTest.directManyChangeString:·gc.count              avgt    5   885.000             counts
PathUtilsBenchmarkTest.directManyChangeString:·gc.time               avgt    5   573.000                 ms
PathUtilsBenchmarkTest.originalManyChangeString                      avgt    5     0.157 ±    0.009   us/op
PathUtilsBenchmarkTest.originalManyChangeString:·gc.alloc.rate       avgt    5  7384.494 ±  418.206  MB/sec
PathUtilsBenchmarkTest.originalManyChangeString:·gc.alloc.rate.norm  avgt    5   152.000 ±    0.001    B/op
PathUtilsBenchmarkTest.originalManyChangeString:·gc.count            avgt    5   610.000             counts
PathUtilsBenchmarkTest.originalManyChangeString:·gc.time             avgt    5  1087.000                 ms

# META-INF/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/.hidden
PathUtilsBenchmarkTest.longNoChangeCase                              avgt    5     1.783 ±    0.030   us/op
PathUtilsBenchmarkTest.longNoChangeCase:·gc.alloc.rate               avgt    5  1882.206 ±   32.076  MB/sec
PathUtilsBenchmarkTest.longNoChangeCase:·gc.alloc.rate.norm          avgt    5   440.000 ±    0.002    B/op
PathUtilsBenchmarkTest.longNoChangeCase:·gc.count                    avgt    5   700.000             counts
PathUtilsBenchmarkTest.longNoChangeCase:·gc.time                     avgt    5   206.000                 ms
PathUtilsBenchmarkTest.directLongNoChangeCase                        avgt    5     1.705 ±    0.016   us/op
PathUtilsBenchmarkTest.directLongNoChangeCase:·gc.alloc.rate         avgt    5  1968.186 ±   18.589  MB/sec
PathUtilsBenchmarkTest.directLongNoChangeCase:·gc.alloc.rate.norm    avgt    5   440.000 ±    0.003    B/op
PathUtilsBenchmarkTest.directLongNoChangeCase:·gc.count              avgt    5   723.000             counts
PathUtilsBenchmarkTest.directLongNoChangeCase:·gc.time               avgt    5   207.000                 ms
PathUtilsBenchmarkTest.originalLongNoChangeCase                      avgt    5     1.849 ±    0.030   us/op
PathUtilsBenchmarkTest.originalLongNoChangeCase:·gc.alloc.rate       avgt    5  5150.149 ±   83.746  MB/sec
PathUtilsBenchmarkTest.originalLongNoChangeCase:·gc.alloc.rate.norm  avgt    5  1248.000 ±    0.003    B/op
PathUtilsBenchmarkTest.originalLongNoChangeCase:·gc.count            avgt    5   632.000             counts
PathUtilsBenchmarkTest.originalLongNoChangeCase:·gc.time             avgt    5   819.000                 ms

# META-INF/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/.
PathUtilsBenchmarkTest.longWorstCase                                 avgt    5     1.732 ±    0.015   us/op
PathUtilsBenchmarkTest.longWorstCase:·gc.alloc.rate                  avgt    5  1903.240 ±   16.433  MB/sec
PathUtilsBenchmarkTest.longWorstCase:·gc.alloc.rate.norm             avgt    5   432.000 ±    0.001    B/op
PathUtilsBenchmarkTest.longWorstCase:·gc.count                       avgt    5   749.000             counts
PathUtilsBenchmarkTest.longWorstCase:·gc.time                        avgt    5   211.000                 ms
PathUtilsBenchmarkTest.directLongWorstCase                           avgt    5     1.700 ±    0.014   us/op
PathUtilsBenchmarkTest.directLongWorstCase:·gc.alloc.rate            avgt    5  1938.810 ±   16.285  MB/sec
PathUtilsBenchmarkTest.directLongWorstCase:·gc.alloc.rate.norm       avgt    5   432.000 ±    0.003    B/op
PathUtilsBenchmarkTest.directLongWorstCase:·gc.count                 avgt    5   729.000             counts
PathUtilsBenchmarkTest.directLongWorstCase:·gc.time                  avgt    5   211.000                 ms
PathUtilsBenchmarkTest.originalLongWorstCase                         avgt    5     1.540 ±    0.014   us/op
PathUtilsBenchmarkTest.originalLongWorstCase:·gc.alloc.rate          avgt    5  6102.714 ±   55.424  MB/sec
PathUtilsBenchmarkTest.originalLongWorstCase:·gc.alloc.rate.norm     avgt    5  1232.000 ±    0.002    B/op
PathUtilsBenchmarkTest.originalLongWorstCase:·gc.count               avgt    5   772.000             counts
PathUtilsBenchmarkTest.originalLongWorstCase:·gc.time                avgt    5   978.000                 ms
```

Broad strokes: there aren't any surprises here. We pay a cost for skipping out early and we pay a price for the ThreadLocal. I think this produces tighter performance in most cases, fewer allocations in most cases, and only seems to cost about 10%-14% in the worst case. In cases that clearly don't need modification this is much faster.

File: pom.xml [hunk 5]
```diff
@@ -35,7 +35,7 @@
     <licenses>
         <license>
             <name>Apache License 2.0</name>
-            <url>http://repository.jboss.org/licenses/apache-2.0.txt</url>
+            <url>http://:y.jboss.org/licenses/apache-2.0.txt</url>
```
dmlloyd: This looks unintentional...

src/main/java/org/jboss/modules/PathUtils.java
```diff
@@ -221,11 +256,59 @@ public static String canonicalize(String path) {
...
+
+    /**
+     * Canonicalize the given path.  Removes all {@code .} and {@code ..} segments from the path.
+     *
+     * @param path the relative or absolute possibly non-canonical path
+     * @return the canonical path
+     */
+    public static String canonicalize(String path) {
+        // is null check here even needed?
+        if (path == null || path.isEmpty()) {
```
dmlloyd: No, I'd skip it and let the NPE fly in that case.

src/main/java/org/jboss/modules/PathUtils.java
```diff
@@ -268,21 +268,41 @@ static String directCanonicalize(String path) {
...
+        }
+        do {
+            // represents a state transitions that will be found via the state machine
+            // but if these patterns aren't contained in the string they can't be found
+            // via the state machine. (we know it already contains at least "." at this point.)
+            if (path.length() > idx + 1 && isSepChar(path.charAt(idx + 1)) || idx > 0 && isSepChar(path.charAt(idx - 1))) {
+                found = true;
+            }
+        } while (idx + 1 < path.length() && (idx = path.indexOf('.', idx + 1)) != -1);
+        if (!found) {
```
dmlloyd: I like this less and less the more complex it gets. I think it misses strings like `/` and `//` and `foo//bar` as well, while duplicating a lot of the work of the state machine.

The right fix is probably to drop all of this stuff and modify the state machine to only call `getBuffer` and initially populate the buffer once it knows that the string will need modifications (e.g. when encountering a `/` in states 1, 2, or 3), and then having the emit routine at the bottom skip emitting unless the buffer is not null. Maybe also replacing the `switch` statements with more optimal `if` trees, which the JIT can profile much better.

chrisruffalo: I tend to agree with you, even at minimal cost doubling up the scan isn't a great idea. I wish we had an early out for cases that 100% don't need it but I don't think it can be done. At one point during my frustration with this I actually had it spit out a list of paths it was trying to canonicalize and it was mostly "application.properties" and "" paths but that could be particular to my use case. I don't know why so that's why I was tripped up looking at what we could do. All of this is coming from isChild or isDirectChild (can't remember) and I don't see any easy optimizations for that use case. I didn't work out at all.

I'll go through the state machine and try and implement those changes if I get some time. I looked at it once more from a reuse perspective and didn't see anything in particular.

If nothing changes I think dropping the added complexity is fine. I was mostly after reducing allocations which seems well handled.

Thanks for going through this. If nothing else it's a fun exercise.

pom.xml
chrisruffalo: Absolutely, I thought I was in another window and didn't realize I made that change.

chrisruffalo:
I spent a few days trying to mess with `switch` vs `if` and didn't get any improvements. I did find a few shortcuts out of the loop and at least one way to seek better. @dmlloyd I know I missed something so it would be really great to get a more considered view of it. I am not 100% sure of the changes I made even with crosschecking.

Anyway I made my focus reducing the number of copies/calls to copy to the array _and_ figuring out a way to detect when the string does not need any editing.

I also made the benchmark an aggregate of all the cases so it shows a better real-world performance. The upside is that it takes less than 3 minutes to run the benchmark even with much longer iterations but the downside is we see less granular results.

The results of this are in the following table:

| Algorithm | Operations (us/op) | Allocation (MB/sec) | Allocation (B/op)  |
|---|---|---|---|
| Current | 0.134 us/op | 1645 MB/sec | 57 B/op |
| Original | 0.286 us/op | 3683 MB/sec | 276 B/op |

My take on this is that the current method is faster overall. The peeking/seeking operation is what makes it faster though and if you remove that it is actually worse (probably due to the ThreadLocal and buffer getting). The current implementation always has less memory pressure. What is notable here is that even with twice the operation speed the memory allocation rate is still about 50%.

Now, one other thing I have thought of is adding a matcher component to the canonicalize method so that we can run it in ways that simulate other filer seperator characters and cross-check other platform configurations.

I was able to emit JFR recordings for the benchmark.

Current:
image
image
image

Original:
image
image
image

chrisruffalo:
@dmlloyd after a frustrating week of troubleshooting it turns out this patch doesn't help in a practical sense. I went through a lot of work to set up jandex indexes in these deployments (cracking third-party jars and inserting them) and got a much greater performance improvement. Something like 100s with the patched modules jar, 30s with jandex indexes.

While this patch might have some merit I can't see allowing this code (my own code, ha) without a practical benefit on large-scale deployments. 

I'm going to close this PR.

PR #302, opened 2023-04-05
refactoring

Fahim-Zaman:
Some refactoring hasdone : 

2 new class created : 
 src/main/java/org/jboss/modules/ConcreteModuleVersionAndDependency.java
 src/main/java/org/jboss/modules/management/Property.java

#Decompose conditional
Performed class : FastCopyHashSet 
function name : add 
line starts : 191 

#Introduce explaining variable
Performed class : FastCopyHashSet
function name : relocate 
line starts : 310 

#Extract method 
Performed class : FastCopyHashSet
function name : remove 
line starts : 520

#Replace conditional with polymorphism
Performed class : JarModuleFinder
function name : findModule 
line starts : 80

#Extract class
Performed class : ObjectProperties ( 2 classes were mentioned in a same file , just split it into two )  
new class name : Property 

#Move method/field
ClassName = ConcreteModuleSpec ( multiple responsibilities were there so created a new class and move some field and functions )
new class name  : ConcreteModuleVersionAndDependency

### DIFF
--- a/src/main/java/org/jboss/modules/ConcreteModuleSpec.java
+++ b/src/main/java/org/jboss/modules/ConcreteModuleSpec.java
@@ -31,34 +31,36 @@ import java.util.Map;
  */
 public final class ConcreteModuleSpec extends ModuleSpec {
 
-    private final String mainClass;
+   
     private final AssertionSetting assertionSetting;
     private final ResourceLoaderSpec[] resourceLoaders;
-    private final DependencySpec[] dependencies;
+ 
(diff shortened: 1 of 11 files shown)


dmlloyd:
Thank you for the submission. However the majority of these refactors are not beneficial and do not serve any particular purpose. Several of them actually affect public API compatibility. In general, I believe that most projects would refuse to accept this kind of PR.

It might be better to try to identify one specific issue and solve that for a first submission, focusing on minimizing the surface area of the change.

PR #305, opened 2023-07-27
[MODULES-439] Add a LoggerFinder implementation to load a finder base…

jamezp:
…d on the boot modules class loader.

https://issues.redhat.com/browse/MODULES-439

### DIFF
--- a/src/main/java/org/jboss/modules/ModuleLoggerFinder.java
+++ b/src/main/java/org/jboss/modules/ModuleLoggerFinder.java
@@ -1,0 +1,130 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2023 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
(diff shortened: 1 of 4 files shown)


src/main/java/org/jboss/modules/ModuleLoggerFinder.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.function.Supplier;
+import java.util.logging.Logger;
+
+/**
+ * A logger finder which attempts to locate a {@link LoggerFinder} on the
+ * {@linkplain java.util.logging.LogManager log managers} class path. If not found a default finder will be used.
+ *
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+public class ModuleLoggerFinder extends LoggerFinder {
```
ropalka: How about making it final class to forbid subclassing?

ropalka (changes requested):

ropalka (approved):

ropalka:
Thanks @jamezp !

PR #307, opened 2023-07-29
[MODULES-441] Lazily load system loggers for cases when they may be c…

jamezp:
…reated before the log manager is configured.

https://issues.redhat.com/browse/MODULES-441

### DIFF
--- a/src/main/java/org/jboss/modules/ModuleLoggerFinder.java
+++ b/src/main/java/org/jboss/modules/ModuleLoggerFinder.java
@@ -20,10 +20,19 @@
 package org.jboss.modules;
 
 import java.lang.System.LoggerFinder;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.Arrays;
+import java.util.Deque;
 import java.util.EnumMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.ResourceBundle;
 import java.util.ServiceLoader;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jboss/modules/ModuleLoggerFinder.java
```diff
@@ -34,64 +43,172 @@
...
-                        if (result == null) {
-                            finder = result = defaultFinder.get();
+                    }
+                    if (finder == null) {
+                        finder = defaultFinder.get();
+                    }
+                } catch (Throwable ignore) {
+                    finder = defaultFinder.get();
+                }
+                // Process the queued loggers and drain the contents to new loggers from the finder found
```
bstansberry, line 110: @jamezp Should there be a common record queue that gets drained? This will end up logging grouped by logger, not by order of record queuing.

jamezp: Yes, that is a good point. I'll fix that.

src/main/java/org/jboss/modules/ModuleLoggerFinder.java
```diff
@@ -34,64 +43,175 @@
...
-
-    public ModuleLoggerFinder() {
-    }
 
+    private static final Map<String, System.Logger> loggers = new ConcurrentHashMap<>();
+    private static final ReentrantLock lock = new ReentrantLock();
+    private static final Deque<SystemLogRecord> messages = new LinkedBlockingDeque<>();
+    private static volatile boolean activated = false;
+    // Guarded by lock
+    private static LoggerFinder finder;
```
ropalka: Although it is guarded by lock it needs to be declared volatile to ensure visibility according to Java Memory model

src/main/java/org/jboss/modules/ModuleLoggerFinder.java
```diff
@@ -34,64 +43,175 @@
...
+                } catch (Throwable ignore) {
+                    finder = defaultFinder.get();
+                }
+                // Process the queued loggers and drain the contents to new loggers from the finder found
+                final Iterator<Map.Entry<String, System.Logger>> iter = loggers.entrySet().iterator();
+                while (iter.hasNext()) {
+                    final Map.Entry<String, System.Logger> entry = iter.next();
+                    final System.Logger currentLogger = entry.getValue();
+                    // Check if we're currently a delegating logger which would need to be replaced with a real backing logger
+                    if (currentLogger instanceof DelegatingSystemLogger) {
```
ropalka: This check is not necessary because only DelegatingSystemLoggers are the values stored in loggers map

src/main/java/org/jboss/modules/ModuleLoggerFinder.java
```diff
@@ -34,64 +43,175 @@
...
+                        final DelegatingSystemLogger delegatingSystemLogger = (DelegatingSystemLogger) currentLogger;
+                        final System.Logger currentDelegate = delegatingSystemLogger.getDelegate();
+                        // If this is a queueing logger, we need to replace with a logger from the finder
+                        if (currentDelegate instanceof QueueingSystemLogger) {
```
ropalka: Thsi check is also not necessary because once we replace the delegate the map entry is being removed on line 126

ropalka (changes requested):

src/main/java/org/jboss/modules/ModuleLoggerFinder.java
jamezp: It's only accessed for read outside the lock, however it should only ever happen once `activated` is `true` which is set within the lock and it, itself is `volatile`. We could add it, I just didn't think the extra read overhead was worth it.

src/main/java/org/jboss/modules/ModuleLoggerFinder.java
jamezp: It feels safe to just check even know we do know. It only happens once at most.

src/main/java/org/jboss/modules/ModuleLoggerFinder.java
jamezp: Same here too. Just feels safer even though it's likely to be the case.

src/main/java/org/jboss/modules/ModuleLoggerFinder.java
bstansberry: Thanks.

ropalka (approved):

PR #314, opened 2023-10-16
[MODULES-445] The name attribute is required in <grant/> in module.xml

hokuda:
Please review this PR.
Thanks in advance.


dmlloyd:
The bug is not in the XML schema but in the parsing: this attribute *should* in fact be optional. I'll present an alternative PR in a moment.

PR #324, opened 2024-02-20
MODULES-446 Print warnings when an overlay directory structure is unreadable

TomasHofman:
Issue: https://issues.redhat.com/browse/JBEAP-26674
Upstream issue: https://issues.redhat.com/browse/MODULES-446
No upstream needed (this is the upstream branch)


dmlloyd (approved):
LGTM. We don't want to go too crazy with logging in this project but I think these initial setup logs are probably helpful.

PR #328, opened 2024-02-23
[MODULES-447] Make the ModuleLoggerFinder.activate() method public. A…

jamezp:
…lso, activate the ModuleLoggerFinder if a threshold is hit.

https://issues.redhat.com/browse/MODULES-447

### DIFF
--- a/src/main/java/org/jboss/modules/ModuleLoggerFinder.java
+++ b/src/main/java/org/jboss/modules/ModuleLoggerFinder.java
@@ -56,6 +56,8 @@ public final class ModuleLoggerFinder extends LoggerFinder {
         }
     };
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/modules/ModuleLoggerFinder.java
```diff
@@ -142,6 +144,24 @@ static void activate(final ClassLoader cl) {
         }
     }
 
+    private static int getMaxLogMessages() {
+        final int dft = 10_000;
```
jamezp, line 148: Please note this is a completely arbitrary number and if there is a better default, please let me know.

ropalka:
Closing - this is not the correct approach to the problem. See discussion at: https://github.com/jboss-modules/jboss-modules/pull/329

PR #329, opened 2024-02-23
[MODULES-447] Make the ModuleLoggerFinder.activate() method public. A…

jamezp:
…lso, activate the ModuleLoggerFinder if a threshold is hit.

https://issues.redhat.com/browse/MODULES-447

Upstream #328

### DIFF
--- a/src/main/java9/org/jboss/modules/ModuleLoggerFinder.java
+++ b/src/main/java9/org/jboss/modules/ModuleLoggerFinder.java
@@ -57,6 +57,7 @@ public final class ModuleLoggerFinder extends LoggerFinder {
         }
     };
 
+    private static final int MAX_LOG_MESSAGES = getMaxLogMessages();
     private static final Map<String, System.Logger> loggers = new ConcurrentHashMap<>();
     private static final ReentrantLock lock = new ReentrantLock();
     private static final Deque<SystemLogRecord> messages = new LinkedBlockingDeque<>();
@@ -140,6 +141,24 @@ public final class ModuleLoggerFinder extends LoggerFinder {
         } finally {
             activated = true;
             lock.unlock();
+        }
+    }
+
+    private static int getMaxLogMessages() {
+        final int dft = 10_000;
+        final String value;
+        if (System.getSecurityManager() == null) {
+            value = System.getProperty("jboss.modules.logger.finder.threshold");
+        } else {
+            value = AccessController.doPrivileged(new PropertyReadAction("jboss.modules.logger.finder.threshold"));
+        }
+        if (value == null) {
+            return dft;
+        }
+        try {
+            return Integer.parseInt(value);
+        } catch (NumberFormatException ignore) {
+            return dft;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java9/org/jboss/modules/ModuleLoggerFinder.java
```diff
@@ -88,7 +89,7 @@ public System.Logger getLogger(final String name, final java.lang.Module module)
      *
      * @param cl the class loader to use
      */
-    static void activate(final ClassLoader cl) {
+    public static void activate(final ClassLoader cl) {
```
dmlloyd: I'd rather not expose this API; what is the use case?

jamezp: There are cases when, for example the WildFly Bootable JAR, we don't use `org.jboss.modules.Main` as the entry point so `activate()` is never invoked. It would be the responsibility of those callers which do not launch through the main entry point to invoke this. I do realize that is a bit odd. I'm definitely open to other options.

https://github.com/wildfly/wildfly-core/pull/5882 is what I'm doing in WildFly Core for now.

dmlloyd: Reflection is probably still best for now. What we need to do is to create an embedding API in JBoss Modules and have a little bit more of a control over the bootstrapping process. Right now most important setup tasks are hard-coded in the `main` method. So ideally, we'd have a bootstrap API with error checking (e.g. "hey you already bootstrapped this once") and then the `main` method would reuse it. Maybe we could do some kind of design session on this in the near future.

jamezp: That makes a lot of sense to me. I'll pull the part of making the API public out, but I do think having the queue length somewhat auto-enable it makes some sense. What do you think of that part @dmlloyd?

dmlloyd: Having a maximum queue length does make sense (we've done similar things in other contexts to avoid queue runaway). But I don't know if activation is the right thing to do in response to hitting the maximum. It seems more like we're just hoping that some large-ish queue depth represents that we've waited "long enough" for things to go through, doesn't it?

jamezp: Yes, the number is completely arbitrary for sure. I guess it would likely be strange to dump a bunch of logs. The idea of activating it was just to have the `System.Logger`'s work rather than loosing logs.

ropalka: The introduction of JBoss Modules bootstrap API is a brilliant idea to solve kind of issues @jamezp is facing @dmlloyd ! When you'll schedule the design meeting please keep me in the loop ;)

ropalka: I just discovered there's a JIRA for it already: https://issues.redhat.com/browse/MODULES-367

jamezp: Goof find @ropalka. I've linked the two JIRA's together.

jamezp:
I've updated this and #328 to both just activate logging if the threshold is hit. I feel like for logging we need to activate, but I also understand it might make the log messages look weird. For example, the date will definitely not be that friendly. However, they are log messages and I don't think we can just discard them. We have to do something.

Another option might be to add the timestamp to the `SimpleLogRecord` and if not activated promptly, default activation would be writing to `System.out` in some formatted pattern.

gaol:
can we merge this if it is what we wanted ? thanks

gaol:
@dmlloyd @ropalka may I ask one of you to check this please ? thanks

dmlloyd:
We are not likely going to use this solution. Discussion is ongoing regarding a potential bootstrap API which would be the replacement. It's not fully decided yet though.

jamezp:
FWIW this is now just a solution which drains the queue if it reaches an arbitrary size. We could make the size configurable with a system property.

petrberan:
Hi @dmlloyd @ropalka is there anything we can do to merge this?

petrberan:
Hi @dmlloyd @ropalka is there anything we can do to merge this?

dmlloyd:
As I said before, this is not the right solution for the problem that was posed. I don't support merging this as it is.

PR #337, opened 2024-03-15
Fix caller sensitivity

dmlloyd:
A caller-sensitive method should never be called from within `jboss-modules`.


ropalka (approved):

ropalka:
Thanks @dmlloyd !

PR #339, opened 2024-03-19
Deprecate `org.jboss.modules.ref`

dmlloyd:
There is now `java.lang.ref.Cleaner`, not to mention equivalent functionality in `smallrye-common` and `wildfly-common` to use instead.


ropalka (approved):

ropalka:
Thanks @dmlloyd !

PR #340, opened 2024-03-28
Allow starting in JPMS mode

dmlloyd:
* Use public lookup when looking for public main method
* Add a missing service requirement for `Policy`

For example: `java -p jboss-modules.jar -m org.jboss.modules ...`

More changes are likely going to be needed to get a full startup of a real application.

### DIFF
--- a/src/main/java/module-info.java
+++ b/src/main/java/module-info.java
@@ -17,11 +17,9 @@
  */
 
 module org.jboss.modules {
-  requires java.desktop;
-  requires java.instrument;
-  requires java.logging;
-  requires java.management;
-  requires java.prefs;
+  // for consumers of this module
+  requires java.se;
+  requires jdk.unsupported;
   exports org.jboss.modules;
   exports org.jboss.modules.filter;
   exports org.jboss.modules.log;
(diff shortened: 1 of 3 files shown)


dmlloyd:
With this last version, it is possible to start up `wildfly-core` using a JPMS module path which includes `jboss-modules` plus `jboss-logmanager` and its transitive dependency set, and have module resolution working properly.

The versions of `jboss-logmanager` which have a module descriptor are not compatible with `wildfly-core` yet, so there is a limiting factor.

src/main/java/module-info.java
```diff
@@ -17,6 +17,9 @@
  */
 
 module org.jboss.modules {
+  // for consumers of this module
+  requires java.se;
```
ropalka, line 21: When we require java.se we can drop all of the following redundant requirements:
 *  requires java.desktop;
 * requires java.instrument;
 * requires java.logging;
 * requires java.management;
 * requires java.prefs;

ropalka (changes requested):

src/main/java/module-info.java
dmlloyd: Good point.

ropalka (approved):

ropalka:
Thanks @dmlloyd !

PR #341, opened 2024-03-28
Allow starting in JPMS mode (2.1.x)

dmlloyd:
* Use public lookup when looking for public main method
* Add a missing service requirement for `Policy`

For example: `java -p jboss-modules.jar -m org.jboss.modules ...`

### DIFF
--- a/src/main/java/module-info.java
+++ b/src/main/java/module-info.java
@@ -17,11 +17,9 @@
  */
 
 module org.jboss.modules {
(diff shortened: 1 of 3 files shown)


dmlloyd:
This is the same as #340 but for the 2.1 branch (since it's only fixed, no new API).

ropalka (approved):

ropalka:
Thanks @dmlloyd !

PR #342, opened 2024-04-02
[MODULES-450] Use java.net.spi.URLStreamHandlerProvider instead of java.net.URLStreamHandlerFactory

ropalka:
https://issues.redhat.com/browse/MODULES-450

Related PR: https://github.com/jbossas/jboss-vfs/pull/57

### DIFF
--- a/src/main/java/org/jboss/modules/ModularURLStreamHandlerProvider.java
+++ b/src/main/java/org/jboss/modules/ModularURLStreamHandlerProvider.java
@@ -20,6 +20,7 @@ package org.jboss.modules;
 
 import java.net.URLStreamHandler;
 import java.net.URLStreamHandlerFactory;
+import java.net.spi.URLStreamHandlerProvider;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.ArrayList;
@@ -33,7 +34,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
  *
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
-final class ModularURLStreamHandlerFactory implements URLStreamHandlerFactory {
+final class ModularURLStreamHandlerProvider extends URLStreamHandlerProvider {
     private static final PrivilegedAction<String> URL_MODULES_LIST_ACTION = new PropertyReadAction("jboss.protocol.handler.modules");
 
     private static final List<Module> modules;
@@ -71,33 +72,39 @@ final class ModularURLStreamHandlerFactory implements URLStreamHandlerFactory {
         modules = list;
     }
 
-    static final ModularURLStreamHandlerFactory INSTANCE = new ModularURLStreamHandlerFactory();
+    static final ModularURLStreamHandlerProvider INSTANCE = new ModularURLStreamHandlerProvider();
 
     static void addHandlerModule(Module module) {
         modules.add(module);
     }
 
-    private ModularURLStreamHandlerFactory() {
+    private ModularURLStreamHandlerProvider() {
     }
 
     private URLStreamHandler locateHandler(final String protocol) {
+        URLStreamHandler handler;
         for (Module module : modules) {
-            ServiceLoader<URLStreamHandlerFactory> loader = module.loadService(URLStreamHandlerFactory.class);
-            for (URLStreamHandlerFactory factory : loader) {
-                try {
-                    final URLStreamHandler handler = factory.createURLStreamHandler(protocol);
-                    if (handler != null) {
-                        return handler;
-                    }
-                } catch (RuntimeException e) {
-                    // ignored
+            handler = loadService(protocol, module, URLStreamHandlerProvider.class);
+            if (handler != null) return handler;
+            // backward compatibility
+            handler = loadService(protocol, module, URLStreamHandlerFactory.class);
+            if (handler != null) return handler;
+        }
+        return "data".equals(protocol) ? DataURLStreamHandler.getInstance() : null;
+    }
+
+    private static <T extends URLStreamHandlerFactory> URLStreamHandler loadService(final String protocol, final Module module, final Class<T> providerType) {
+        final ServiceLoader<T> loader = module.loadService(providerType);
+        for (URLStreamHandlerFactory factory : loader) {
+            try {
+                final URLStreamHandler handler = factory.createURLStreamHandler(protocol);
+                if (handler != null) {
+                    return handler;
                 }
+            } catch (RuntimeException e) {
+                // ignored
             }
         }
-        if (protocol.equals("data")) {
-            return DataURLStreamHandler.getInstance();
-        }
-
         return null;
     }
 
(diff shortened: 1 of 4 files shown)


ropalka:
I know this change isn't backward compatible @dmlloyd but it goes hand in hand with JBoss VFS change to keep it functional. Or do you consider it a problem?

dmlloyd:
I think it makes sense but I have to review this `java.net` API.

ropalka:
OK, this change is related to this change in JDK code base: https://github.com/openjdk/jdk11u/commit/b7d07021882

src/main/java/org/jboss/modules/ModularURLStreamHandlerProvider.java
```diff
@@ -71,19 +71,19 @@ final class ModularURLStreamHandlerFactory implements URLStreamHandlerFactory {
...
-    private ModularURLStreamHandlerFactory() {
+    private ModularURLStreamHandlerProvider() {
     }
 
     private URLStreamHandler locateHandler(final String protocol) {
         for (Module module : modules) {
-            ServiceLoader<URLStreamHandlerFactory> loader = module.loadService(URLStreamHandlerFactory.class);
-            for (URLStreamHandlerFactory factory : loader) {
+            ServiceLoader<URLStreamHandlerProvider> loader = module.loadService(URLStreamHandlerProvider.class);
+            for (URLStreamHandlerProvider factory : loader) {
```
dmlloyd: I think we should load *both* service types. The existing API contract we give supports loading `URLStreamHandlerFactory` so dropping that behavior would be an incompatible change; I think we can prefer the new `URLStreamHandlerProvider` and then fall back to the old `URLStreamHandlerFactory`.

src/main/java/org/jboss/modules/Module.java
```diff
@@ -1217,13 +1219,32 @@ public static void registerContentHandlerFactoryModule(Module module) {
...
+     * access is not granted, a {@code SecurityException} will be thrown.
+     *
+     * @param module the module to add
+     */
+    public static void registerURLStreamHandlerProviderModule(Module module) {
+        final SecurityManager sm = System.getSecurityManager();
+        if (sm != null) {
+            sm.checkPermission(ADD_URL_STREAM_HANDLER_PROVIDER);
+        }
+        ModularURLStreamHandlerProvider.addHandlerModule(module);
```
dmlloyd: I don't think we need to change this method. We can retain the original method and have it support both interfaces as above.

src/main/java/org/jboss/modules/Module.java
```diff
@@ -212,6 +213,7 @@ public static ModulesPrivateAccess getPrivateAccess() {
     private static final RuntimePermission ACCESS_MODULE_LOGGER;
     private static final RuntimePermission ADD_CONTENT_HANDLER_FACTORY;
     private static final RuntimePermission ADD_URL_STREAM_HANDLER_FACTORY;
+    private static final RuntimePermission ADD_URL_STREAM_HANDLER_PROVIDER;
```
dmlloyd: We should not change the permission, especially since the `SecurityManager` is on its way towards obsolescence. We should keep the old one until we don't need any permissions anymore.

src/main/java/org/jboss/modules/ModularURLStreamHandlerProvider.java
```diff
@@ -33,7 +33,7 @@
  *
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
-final class ModularURLStreamHandlerFactory implements URLStreamHandlerFactory {
+final class ModularURLStreamHandlerProvider extends URLStreamHandlerProvider {
```
dmlloyd, line 37: This class rename is OK. :+1:

src/main/java/org/jboss/modules/Module.java
```diff
@@ -113,7 +114,7 @@ public final class Module {
         AccessController.doPrivileged(new PrivilegedAction<Void>() {
             public Void run() {
                 try {
-                    URL.setURLStreamHandlerFactory(ModularURLStreamHandlerFactory.INSTANCE);
+                    URL.setURLStreamHandlerFactory(ModularURLStreamHandlerProvider.INSTANCE);
```
dmlloyd, line 116: Maybe we should eliminate this (and the similar block below), and replace it with `provides URLStreamHandlerProvider with ModularURLStreamHandlerProvider` (and corresponding `META-INF/services` files) in the module-info? It would require these classes being made public, however we could move them to the non-exported `_private` subpackage in this case.

The advantage of doing this is that it would be one less step for the bootstrap API to deal with. However we could make this change in a follow-up pull request after more discussion if needed.

dmlloyd (changes requested):
I think this change is on the right track but I think we need to do it slightly differently (see comments).

Thanks for your patience!

src/main/java/org/jboss/modules/ModularURLStreamHandlerProvider.java
ropalka: I thought so.

src/main/java/org/jboss/modules/Module.java
ropalka: OK.

src/main/java/org/jboss/modules/Module.java
ropalka: Makes sense.

src/main/java/org/jboss/modules/ModularURLStreamHandlerProvider.java
ropalka: Thanks!

src/main/java/org/jboss/modules/Module.java
ropalka: I created https://issues.redhat.com/browse/MODULES-451 as follow up of this proposal.

ropalka:
I'm not able to reproduce the org.jboss.modules.MavenResource2Test.testDefaultRepositories failure locally :(
Seems it is something CI environment specific @dmlloyd  ? My configuration is:

>$ uname -a
Linux blackhole 6.7.11-200.fc39.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Mar 27 16:50:39 UTC 2024 x86_64 GNU/Linux
>$ java -version
openjdk version "11.0.22" 2024-01-16
OpenJDK Runtime Environment Temurin-11.0.22+7 (build 11.0.22+7)
OpenJDK 64-Bit Server VM Temurin-11.0.22+7 (build 11.0.22+7, mixed mode)

dmlloyd:
To reproduce the problem, you will likely need to install and build with JDK 21.

ropalka:
JDK21 works for me:

>$ java -version
openjdk version "21.0.2" 2024-01-16 LTS
OpenJDK Runtime Environment Temurin-21.0.2+13 (build 21.0.2+13-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.2+13 (build 21.0.2+13-LTS, mixed mode, sharing)

dmlloyd:
It looks like it was a transient failure in the CI environment.

dmlloyd (approved):

ropalka:
Thanks for review @dmlloyd !

PR #347, opened 2024-04-09
[MODULES-450] Use java.net.spi.URLStreamHandlerProvider instead of java.net.URLStreamHandlerFactory 

ropalka:
https://issues.redhat.com/browse/MODULES-450

Downstream PR: https://github.com/jboss-modules/jboss-modules/pull/342

### DIFF
--- a/src/main/java/org/jboss/modules/ref/package-info.java
+++ b/src/main/java/org/jboss/modules/ref/package-info.java
@@ -23,5 +23,4 @@
  * @deprecated Use {@link java.lang.ref.Cleaner} or one of the reference types from a support library such as
  *      {@code io.smallrye.common:smallrye-common-ref} instead.
  */
-@Deprecated
 package org.jboss.modules.ref;
--- a/src/main/java/module-info.java
+++ b/src/main/java/module-info.java
@@ -28,6 +28,8 @@ module org.jboss.modules {
   exports org.jboss.modules.ref;
   exports org.jboss.modules.security;
   exports org.jboss.modules.xml;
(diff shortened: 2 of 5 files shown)


src/main/java/org/jboss/modules/ref/package-info.java
```diff
@@ -23,5 +23,4 @@
  * @deprecated Use {@link java.lang.ref.Cleaner} or one of the reference types from a support library such as
  *      {@code io.smallrye.common:smallrye-common-ref} instead.
  */
-@Deprecated
-package org.jboss.modules.ref;
\ No newline at end of file
+package org.jboss.modules.ref;
```
dmlloyd, line 26: Did you mean to un-deprecate this?

ropalka: No, when building the 2.x branch locally I'm getting build failure if @deprecated annotation is used on package name @dmlloyd :

---
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/opalka/git/redhat/PUBLIC/jboss-modules/src/main/java/org/jboss/modules/ref/package-info.java:[27,1] modifier deprecated not allowed here
[INFO] 1 error
---

dmlloyd: I guess it must be due to having an earlier JDK target. That's fine then!

dmlloyd (approved):

ropalka:
Thanks for review @dmlloyd !

PR #348, opened 2024-04-10
Refactoring - unifying service loading mechanism in Main. Always use Module.loadService() instead of ServiceLoader.load()

ropalka:


dmlloyd (approved):

ropalka:
Thanks for review @dmlloyd !

PR #352, opened 2024-11-19
[MODULES-453] Ignore setting up the security policy in JVM's where th…

jamezp:
…e security manager support is disabled. Fail if attempting to enable the security manager in JVM's where the security manager is not supported.

https://issues.redhat.com/browse/MODULES-453

### DIFF
--- a/src/main/java/org/jboss/modules/Main.java
+++ b/src/main/java/org/jboss/modules/Main.java
@@ -44,7 +44,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Properties;
 import java.util.ServiceConfigurationError;
-import java.util.ServiceLoader;
 import java.util.logging.LogManager;
 
 import java.util.jar.Attributes;
@@ -280,6 +279,12 @@ public final class Main {
                 System.exit(1);
             }
         }
+        int actualVersion = Runtime.version().feature();
+
+        if (actualVersion >= 24 && (secMgrModule != null || defaultSecMgr)) {
+            System.err.printf("The security manager is not supported in Java %d.", Runtime.version().feature());
+            System.exit(1);
+        }
 
         if (deps != null && ! classDefined && ! classpathDefined) {
             System.err.println("-deps may only be specified when -cp/-classpath or -class is in use");
@@ -358,7 +363,6 @@ public final class Main {
             System.exit(1);
             return;
         }
-        int actualVersion = Runtime.version().feature();
         final String requireJavaVersion = module.getProperty("jboss.require-java-version", Integer.toString(actualVersion));
         final Pattern versionPattern = Pattern.compile("(?:1\\.)?(\\d+)");
         final Matcher requireMatcher = versionPattern.matcher(requireJavaVersion);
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/modules/Main.java
```diff
@@ -370,50 +378,51 @@ public static void main(String[] args) throws Throwable {
...
 
-        // at this point, having a security manager already installed will prevent correct operation.
-
-        final SecurityManager existingSecMgr = System.getSecurityManager();
-        if (existingSecMgr != null) {
-            System.err.println("An existing security manager was detected.  You must use the -secmgr switch to start with a security manager.");
-            System.exit(1);
-            return; // not reached
-        }
+        if (SECURITY_MANAGER_SUPPORTED) {
```
dmlloyd: I think it would be marginally clearer to use the `actualVersion` variable above, instead of a constant field which is only going to be used within this method but which persists on the class forever.

src/main/java/org/jboss/modules/Main.java
```diff
@@ -248,6 +252,10 @@ public static void main(String[] args) throws Throwable {
                         }
                         defaultSecMgr = true;
                     } else if ("-secmgrmodule".equals(arg)) {
+                        if (!SECURITY_MANAGER_SUPPORTED) {
+                            System.err.printf("The security manager is not supported in Java %d.", Runtime.version().feature());
+                            System.exit(1);
+                        }
```
dmlloyd: Since these checks are repeated, you could instead just check below e.g. `if (actualVersion >= 24 && (secMgrModule != null || defaultSegMgr)) { .. print the error .. }`.

.github/workflows/ci.yml
```diff
@@ -45,3 +46,10 @@ jobs:
       with:
         name: surefire-${{ matrix.jdk-distribution }}-${{ matrix.os }}
         path: '**/surefire-reports/*.txt'
+    - name: Run Tests on Java 24
+      run: mvn -ntp -U -B -fae clean install "-Djvm=${{env.JAVA_HOME_24_X64}}/bin/java"
+    - uses: actions/upload-artifact@v4
+      if: failure()
+      with:
+        name: surefire-${{ matrix.jdk-distribution }}-${{ matrix.os }}-24
+        path: '**/surefire-reports/*.txt'
```
dmlloyd: We need to remember to update this once 24 is finalized, and again when the next LTS is out (to remove this extra test run).

jamezp: Yeah, I guess the version name in there is probably a bit much.

src/main/java/org/jboss/modules/Main.java
jamezp: That makes sense.

.github/workflows/ci.yml
```diff
@@ -45,3 +45,27 @@ jobs:
...
+        with:
+          distribution: ${{ matrix.jdk-distribution }}
+          java-version: 24-ea
+      - name: Run Tests
+        run: mvn -ntp -U -B -fae clean install "-Djvm=${{env.JAVA_HOME_24_X64}}/in/java"
+      - uses: actions/upload-artifact@v4
+        if: failure()
+        with:
+          name: surefire-${{ matrix.jdk-distribution }}-${{ matrix.os }}-latest
+          path: '**/surefire-reports/*'
```
jamezp: FWIW I'm not sure if this does much good as the tests pass on Java 24 without my changes since the entry point is not tested.

dmlloyd: I think it's probably fine to drop it then if you want. Extra CI always comes at a cost.

src/main/java/org/jboss/modules/Main.java
```diff
@@ -76,6 +75,7 @@ public final class Main {
     }
 
     private static final String[] NO_STRINGS = new String[0];
+    private static final int JVM_VERSION = Runtime.version().feature();
```
dmlloyd: I would reuse the `actualVersion` variable on line ~366 (move it up a little if needed) and drop the constant field.

jamezp: My bad,  didn't even see that :). Fixed now.

dmlloyd (approved):

PR #354, opened 2024-11-19
[MODULES-453] Add deprecation warnings for Security Manager related command line arguments and skip providers setup on JDK24+

ropalka:
https://issues.redhat.com/browse/MODULES-453

### DIFF
--- a/src/main/java/org/jboss/modules/Main.java
+++ b/src/main/java/org/jboss/modules/Main.java
@@ -238,6 +238,7 @@ public final class Main {
                         }
                         classDefined = true;
                     } else if ("-secmgr".equals(arg)) {
+                        System.err.println("WARNING: -secmgr is deprecated and may be removed in a future release");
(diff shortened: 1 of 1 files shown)


bstansberry:
I consider writing warning messages to stderr to be a breaking change.

dmlloyd (changes requested):
I don't think this is correct. Security provider registration should still be supported in 24+ IIRC, and is not related to the security manager removal.

bstansberry:
My preference is that deprecating the SM in visible-at-runtime ways (i.e. not @Deprecated) be left to Java SE or to higher level applications like WildFly / EAP. It will be very hard to do anything coherent if libraries are trying to drive end-user behavior.

ropalka:
From my understanding both proposed changes
- added deprecation warning for security related options
-  disabling security provider registrations

are not valid.

Issue #736, opened 2024-02-23
[MODULES-447] The ModuleLoggerFinder should activate to avoid a runaway queue


dmlloyd:
**[James Perkins](https://issues.redhat.com/secure/ViewProfile.jspa?name=jamezp)** opened **[MODULES-447](https://issues.redhat.com/browse/MODULES-447?redirect=false)** and commented

-The `ModuleLoggerFinder.activate()` method should be made public so that it can be invoked from other entry points like the WildFly Bootable JAR.-

MODULES-367 should introduce an API to handle the activation. However, if activation is not done something should happen to the message queue so that it does not fill up.

As part of this a threshold of messages should used and if that threshold is hit, activation should happen delegating to JUL.

---

**Issue Links:**
- [WFCORE-6721](https://issues.redhat.com/browse/WFCORE-6721) Do not use reflection to activate System.Logger for Bootable JARs and Embedded servers
 (_**"blocks"**_)
- [JBEAP-26747](https://issues.redhat.com/browse/JBEAP-26747) (8.0.z) MODULES-447 - The ModuleLoggerFinder should activate to avoid a runaway queue
 (_**"is incorporated by"**_)
- [JBEAP-26746](https://issues.redhat.com/browse/JBEAP-26746) (7.4.z) MODULES-447 - The ModuleLoggerFinder should activate to avoid a runaway queue
 (_**"is incorporated by"**_)
- [WFCORE-6712](https://issues.redhat.com/browse/WFCORE-6712) Bootable JAR deployments cannot use the System.Logger

- [JBEAP-26691](https://issues.redhat.com/browse/JBEAP-26691) (8.0.z) Bootable JAR deployments cannot use the System.Logger

- [JBEAP-26689](https://issues.redhat.com/browse/JBEAP-26689) (7.4.z) Bootable JAR deployments cannot use the System.Logger

- [MODULES-367](https://issues.redhat.com/browse/MODULES-367) Bootstrap API

- [MODULES-441](https://issues.redhat.com/browse/MODULES-441) Lazily initialize System.Logger's in the ModuleLoggerFinder
 (_**"is triggered by"**_)

dmlloyd:
**[James Perkins](https://issues.redhat.com/secure/ViewProfile.jspa?name=jamezp)** commented

I've changed the priority to critical because this could lead to an OOME if the message queue gets too large.

PR #791, opened 2025-04-11
OOME if the message queue gets too large.

YJP-D:
#736 
The resource file [META-INF/services/java.lang.System$LoggerFinder]  in jar will cause a OOM error.

### DIFF
--- a/src/main/java/org/jboss/modules/ModuleLoggerFinder.java
+++ b/src/main/java/org/jboss/modules/ModuleLoggerFinder.java
@@ -249,6 +249,8 @@ public final class ModuleLoggerFinder extends LoggerFinder {
     private static class QueueingSystemLogger implements System.Logger {
         private static final System.Logger.Level DEFAULT_LEVEL;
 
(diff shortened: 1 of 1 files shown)


dmlloyd:
Note that the expectation is that you'll have some other system log provider installed at some point. If you do not, then yes, the queue will overflow. I agree that it makes sense to have a cap though.

dmlloyd (approved):

30035122:
@dmlloyd 
I have a question. Could you let me know when the changes from the merge request will be released?
I’m looking forward to using the released version.


CC　@YJP-D

dmlloyd:
I think I can probably do a release early next week if that's OK.

30035122:
@dmlloyd
Thank you for your reply.I appreciate that you’ve scheduled the release.

CC　@YJP-D

dmlloyd:
See #795. Still waiting for a Nexus issue to be resolved so I can cut the release.

dmlloyd:
It looks like a release may still not be possible for the next few weeks.

